PR #4, opened 2022-09-01
Adds register screen

andrewhamilton0:
Adds register screen, with buttons letting you go between activities. Also adds visibility buttons letting you see password. Edits LoginActivity to make it portrait only and able to use different screen sizes by using guidelines.


app/src/main/java/com/andrew/tasky/presentation/login/LoginActivity.kt
```diff
@@ -0,0 +1,71 @@
...
+
+
+        val emailAddress = findViewById<EditText>(R.id.emailAddressEditText)
+        val emailAddressCheckBox = findViewById<ImageView>(R.id.emailAddressCheckBox)
+        val password = findViewById<EditText>(R.id.passwordEditText)
+        val passwordVisibilityButton = findViewById<ImageButton>(R.id.passwordVisibilityButton)
+        var passwordCursorPlace = Int
+        var passwordVisibility = false
+        val loginButton = findViewById<Button>(R.id.loginButton)
+        val signUpButton = findViewById<TextView>(R.id.signUpButton)
```
philipplackner, line 29: I'd recommend to take a look at ViewBinding to get rid of all these nullable views, much more convenient and quick to learn :)

app/src/main/java/com/andrew/tasky/presentation/login/LoginActivity.kt
```diff
@@ -0,0 +1,71 @@
...
+
+
+
+
+
+
```
philipplackner, line 35: Any reason for all these empty lines? I'd use an empty line to signalize a context switch in code, but not more than that - that makes it a bit hard to quickly fly over this class

app/src/main/java/com/andrew/tasky/presentation/login/LoginActivity.kt
```diff
@@ -0,0 +1,71 @@
...
+        password.addTextChangedListener(){
+
+        }
+
+
+        passwordVisibilityButton.setOnClickListener(){
+            if(!passwordVisibility){
+                passwordVisibilityButton.setBackgroundResource(R.drawable.ic_baseline_visibility_24)
+                passwordVisibility = true
```
philipplackner, line 56: This state would get lost on a configuration change such as a screen rotation. Better use a ViewModel for such values.

app/src/main/java/com/andrew/tasky/presentation/register/RegisterActivity.kt
```diff
@@ -0,0 +1,64 @@
...
+            if(!passwordVisibility){
+                passwordVisibilityButton.setBackgroundResource(R.drawable.ic_baseline_visibility_24)
+                passwordVisibility = true
+                password.transformationMethod = HideReturnsTransformationMethod.getInstance()
+            }
+            else{
+                passwordVisibilityButton.setBackgroundResource(R.drawable.ic_baseline_visibility_off_24)
+                passwordVisibility = false
+                password.transformationMethod = PasswordTransformationMethod.getInstance()
+            }
```
philipplackner, line 60: It would make sense to make this password textfield a custom view to be able to reuse it on different screens to minimize the duplicate code here. Imagine, the requirements for this text field change in a real scenario, then you'd need to update all screens where you use that text field instead of just one class.
If custom views are new to you, I'll be happy to help you out :)

philipplackner (approved):

PR #6, opened 2022-09-25
Adds edit mode, a few more icons, time and date dialog box, and bundles

andrewhamilton0:
Adds edit mode, a few more icons, time and date dialog box, and bundles


app/src/main/java/com/andrew/tasky/presentation/task_detail/TaskDetailFragment.kt
```diff
@@ -4,29 +4,198 @@ import androidx.lifecycle.ViewModelProvider
...
+        }
+
+        binding.date.setOnClickListener {
+            datePickerFragment()
+        }
+    }
+
+    private fun subscribeToObservables() {
+        lifecycleScope.launchWhenStarted {
+            viewModel.selectedDate.collectLatest {
```
philipplackner, line 93: 1. In fragments, always use `viewLifecycleOwner.lifecycleScope` which is tied to the lifecycle of what's visible (you only want to update views when they are actually active)
2. It is recommended to use the `repeatOnLifecycle` API to collect Flows in the UI. Not the end of the world for StateFlows like here, but as soon as you deal with events (for example coming from a SharedFlow), these can get lost if you're unlucky.
The reason is that lifecycleScope is cancelled when `onStop()` is reached. Therefore, if an event comes in between `onStop()` and `onDestroy()`, it would be lost since there are no active observers.

app/src/main/java/com/andrew/tasky/presentation/task_detail/TaskDetailFragment.kt
```diff
@@ -4,29 +4,198 @@ import androidx.lifecycle.ViewModelProvider
...
+            }
+        }
+        lifecycleScope.launchWhenStarted {
+            viewModel.selectedTime.collectLatest {
+                binding.time.text = it
+            }
+        }
+    }
+
+    private fun timePickerFragment(){
```
philipplackner, line 104: Naming recommendation: `showTimePickerFragment()` since that's what it does right? 😄 
Same for datePickerFragment() below

app/src/main/java/com/andrew/tasky/presentation/task_detail/TaskDetailFragment.kt
```diff
@@ -4,29 +4,198 @@ import androidx.lifecycle.ViewModelProvider
...
+        binding.descriptionTextView.isClickable = true
+
+        binding.time.isClickable = true
+        binding.date.isClickable = true
+
+        binding.reminderButton.isClickable = true
+        binding.reminderButton.isVisible = true
+    }
+
+    private fun nonEditMode(){
```
philipplackner, line 161: This function is redundant, you can instead create one like this:
```
private fun setEditMode(isEditing: Boolean){
        binding.currentDateTextView.text = getString(R.string.edit_task_text)

        binding.saveButton.isVisible = isEditing
        binding.saveButton.isClickable = isEditing

        binding.editButton.isVisible = !isEditing
        binding.editButton.isClickable = !isEditing

        binding.editTitleButton.isVisible = isEditing
        binding.editTitleButton.isClickable = isEditing
        binding.taskTitle.isClickable = isEditing

        binding.editDescriptionButton.isVisible = isEditing
        binding.editDescriptionButton.isClickable = isEditing
        binding.descriptionTextView.isClickable = isEditing

        binding.time.isClickable = isEditing
        binding.date.isClickable = isEditing

        binding.reminderButton.isClickable = isEditing
        binding.reminderButton.isVisible = isEditing
    }
```
and you use that passed boolean to update your views. Considering the feedback from above, you can then simply call this in the observer of your `isInEditMode´ Flow from the ViewModel

app/src/main/java/com/andrew/tasky/presentation/task_detail/TaskDetailViewModel.kt
```diff
@@ -1,6 +1,37 @@
...
+import com.andrew.tasky.util.TaskType
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import java.text.SimpleDateFormat
+import java.util.*
 
 class TaskDetailViewModel : ViewModel() {
+
+    private val calendar = Calendar.getInstance()
+    private val _selectedDate = MutableStateFlow(SimpleDateFormat("MMM dd yyyy", Locale.ENGLISH).format((calendar.time)))
```
philipplackner, line 13: I'd recommend to take a look at the Java Date Time API instead of `SimpleDateFormat` which offers much more cool functionality (like calculating with dates, adding/subtracting days, months etc.) and it's thread-safe (which `SimpleDateFormat` is not: https://www.callicoder.com/java-simpledateformat-thread-safety-issues/#:~:text=One%20of%20the%20most%20important,environments%20if%20not%20used%20properly.)

philipplackner, line 13: I'd recommend to take a look at the Java Date Time API instead of `SimpleDateFormat` which is a lot easier to deal with and it's thread-safe (which `SimpleDateFormat` is not: https://www.callicoder.com/java-simpledateformat-thread-safety-issues/#:~:text=One%20of%20the%20most%20important,environments%20if%20not%20used%20properly.)

app/src/main/java/com/andrew/tasky/presentation/task_detail/TaskDetailViewModel.kt
```diff
@@ -1,6 +1,37 @@
...
+    val selectedDate = _selectedDate.asStateFlow()
+
+    private val _selectedTime = MutableStateFlow(SimpleDateFormat("hh:mm a", Locale.ENGLISH).format((calendar.time)))
+    val selectedTime = _selectedTime.asStateFlow()
+
+    private var editMode: Boolean = false
```
philipplackner, line 19: As mentioned above, this needs to be a state in form of a `StateFlow`.
Every value that can change over time and is used in the UI is considered UI state.

philipplackner (approved):

app/src/main/java/com/andrew/tasky/presentation/task_detail/TaskDetailFragment.kt
philipplackner: https://medium.com/androiddevelopers/a-safer-way-to-collect-flows-from-android-uis-23080b1f8bda

PR #7, opened 2022-09-27
Cleans code up and adds title and description bundle to bring them to…

andrewhamilton0:
Cleans code up and adds title and description bundle to bring them to EditFragment.kt and back to AgendaItemDetailFragment. Also renames a few items to be more clear.


app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -0,0 +1,257 @@
...
+            editTitleButton.setOnClickListener {
+                editFragment(EditType.TITLE)
+            }
+
+            editDescriptionButton.setOnClickListener {
+                editFragment(EditType.DESCRIPTION)
+            }
+
+            descriptionTextView.setOnClickListener {
+                editFragment(EditType.DESCRIPTION)
```
philipplackner, line 132: Similar as for `showTimePickerFragment()` I'd recommend calling this something like `navigateToEditFragment()`

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailViewModel.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+class AgendaItemDetailViewModel : ViewModel() {
+
+    private val _isInEditMode = MutableStateFlow(false)
+    val isInEditMode = _isInEditMode.asStateFlow()
+    fun setEditMode(x: Boolean){
```
philipplackner, line 13: I'd avoid using variable names like `x`. Even though, in this case it's obvious, it's always better to have a name that tells someone what it stands for like `isEditing` which also makes it easier to scan such code without first needing to think "Uh what was that for again"

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/DatePickerFragment.kt
```diff
@@ -1,32 +1,31 @@
...
 
-    private val calendar = Calendar.getInstance()
+    private val currentDateTime = LocalDateTime.now()
 
     override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
-        val year = calendar.get(Calendar.YEAR)
-        val month = calendar.get(Calendar.MONTH)
-        val day = calendar.get(Calendar.DAY_OF_MONTH)
+        val year = currentDateTime.year
+        val month = currentDateTime.monthValue-1
```
philipplackner, line 19: What's the reason for `-1`? 
For these things I personally like to add a comment explaining why, since even if you're working on this project alone, you'll ask yourself why you subtracted 1 here if you come back to this after a few weeks.

Generally, whenever someone could ask themselves "Why is this happening?" a comment is great. If you add a comment that explains WHAT happens, it's usually a sign that you could improve the readability of the code to get rid of the comment :)

philipplackner (approved):

PR #8, opened 2022-09-30
Adds agendaItemAdapter and AgendaItem

andrewhamilton0:
Adds agendaItemAdapter and AgendaItem to make a recyclerview of agenda items on agenda screen

### DIFF
--- a/app/src/main/java/com/andrew/tasky/presentation/adapter/AgendaItemAdapter.kt
+++ b/app/src/main/java/com/andrew/tasky/presentation/adapter/AgendaItemAdapter.kt
@@ -1,0 +1,64 @@
+package com.andrew.tasky.presentation.adapter
+
+import android.graphics.Color
+import android.graphics.Paint
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import com.andrew.tasky.R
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/andrew/tasky/presentation/adapter/AgendaItemAdapter.kt
```diff
@@ -0,0 +1,62 @@
...
+        val layoutInflater = LayoutInflater.from(parent.context)
+        val binding = ItemAgendaBinding.inflate(layoutInflater, parent, false)
+        return AgendaItemViewHolder(binding)
+    }
+
+    override fun onBindViewHolder(holder: AgendaItemViewHolder, position: Int) {
+        holder.binding.apply {
+            agendaItemTitle.text = agendaItems[position].title
+            agendaItemDescription.text = agendaItems[position].description
+            agendaItemDate.text = agendaItems[position].fromDate + ", " + agendaItems[position].fromTime
```
philipplackner, line 29: In general, I always keep times and dates as `LocalDateTime` in my models. That has the advantage that it's up to the UI how to display it. You might want to have different format patterns on different screens which doesn't work if you format once and attach the date as String to the model :)

philipplackner (approved):

PR #9, opened 2022-10-04
Adds delete confirmation dialog and reminder time options menu popup

andrewhamilton0:
Adds delete confirmation dialog and reminder time options menu popup. Also changes AgendaItem.kt to take LocalDateAndTime instead of a string for start date and end date

### DIFF
--- a/app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
+++ b/app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
@@ -1,9 +1,11 @@
 package com.andrew.tasky.presentation.agenda_item_detail
 
+import android.app.Dialog
 import androidx.lifecycle.ViewModelProvider
 import android.os.Bundle
+import android.view.*
+import android.widget.PopupMenu
 import androidx.fragment.app.Fragment
-import android.view.View
 import androidx.core.view.isVisible
 import androidx.fragment.app.setFragmentResult
 import androidx.fragment.app.setFragmentResultListener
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -240,10 +262,77 @@ class AgendaItemDetailFragment : Fragment(R.layout.fragment_agenda_item_detail)
             descriptionTextView.isClickable = isEditing
 
             time.isClickable = isEditing
+            timeButton.isVisible = isEditing
+            timeButton.isClickable=isEditing
```
philipplackner, line 266: I'm not sure if `isClickable` has an influence.. I think this refers to showing a ripple effect. Better use `timeButton.isEnabled = isEditing` to disable/enable it

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -240,10 +262,77 @@ class AgendaItemDetailFragment : Fragment(R.layout.fragment_agenda_item_detail)
...
+                    }
+                    else -> true
+                }
+            }
+        }
+        popupMenu.show()
+    }
+
+    private fun showDeleteDialog(){
+        val deleteDialog = Dialog(requireActivity())
```
philipplackner, line 312: I'd recommend to make this a `DialogFragment` instead (and put its logic into its own subclass, so you can easily reuse it). That's effectively a fragment with the same lifecycle a fragment has, you can use ViewModels there and it's also not dismissed on screen rotations like the normal Dialog

philipplackner (approved):

PR #10, opened 2022-10-07
10/06/2022 adds mini calendar

andrewhamilton0:
Adds miniCalendar feature with selectable dates. Added FAB on agenda screen with popup menu. Removed the test buttons I made for adding agenda items. Added logout button. Added popup menu for AgendaItemAdapter.kt

### DIFF
--- a/app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
+++ b/app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
@@ -13,6 +13,7 @@ import com.andrew.tasky.data.AgendaItem
 import com.andrew.tasky.databinding.FragmentAgendaBinding
 import com.andrew.tasky.presentation.adapter.AgendaItemAdapter
 import com.andrew.tasky.util.AgendaItemType
+import java.time.LocalDateTime
 
 class AgendaFragment : Fragment(R.layout.fragment_agenda) {
 
@@ -29,13 +30,15 @@ class AgendaFragment : Fragment(R.layout.fragment_agenda) {
 
         var agendaItemList = mutableListOf(
             AgendaItem(AgendaItemType.EVENT, false, "Meeting",
-                "Project meeting with Abby", "3:00", "Mar 5"),
+                "Project meeting with Abby", LocalDateTime.of(2022, 10, 3, 8,22),
+                LocalDateTime.of(2022, 10, 3, 8,32)),
             AgendaItem(AgendaItemType.REMINDER, false, "Take out trash",
-                "Take trash from inside to outside", "4:00", "Mar 5"),
+                "Take trash from inside to outside", LocalDateTime.of(2022, 10, 3, 8,37)),
             AgendaItem(AgendaItemType.TASK, true, "Code",
-                "Finish coding lesson", "5:00", "Mar 5"),
+                "Finish coding lesson", LocalDateTime.of(2022, 10, 3, 9,22)),
             AgendaItem(AgendaItemType.EVENT, false, "Run",
-                "Go for an evening run", "6:00", "Mar 5"),
+                "Go for an evening run", LocalDateTime.of(2022, 10, 3, 12,22),
+                LocalDateTime.of(2022, 10, 15, 14,22)),
         )
 
         val adapter = AgendaItemAdapter(agendaItemList)
--- a/app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
+++ b/app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
@@ -1,67 +1,279 @@
 package com.andrew.tasky.presentation.agenda
 
+import android.graphics.Color
 import androidx.lifecycle.ViewModelProvider
 import android.os.Bundle
 import androidx.fragment.app.Fragment
 import android.view.View
-import androidx.core.os.bundleOf
+import android.widget.PopupMenu
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
 import androidx.navigation.NavController
 import androidx.navigation.Navigation
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.andrew.tasky.R
 import com.andrew.tasky.data.AgendaItem
 import com.andrew.tasky.databinding.FragmentAgendaBinding
+import com.andrew.tasky.databinding.ItemMiniCalendarDayBinding
 import com.andrew.tasky.presentation.adapter.AgendaItemAdapter
 import com.andrew.tasky.util.AgendaItemType
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.launch
+import java.time.LocalDate
 import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
 
 class AgendaFragment : Fragment(R.layout.fragment_agenda) {
 
     private lateinit var navController: NavController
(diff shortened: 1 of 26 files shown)


app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
```diff
@@ -1,64 +1,279 @@
...
+            ),
         )
 
         val adapter = AgendaItemAdapter(agendaItemList)
 
-        binding.agendaItemRecyclerView.adapter = adapter
-        binding.agendaItemRecyclerView.layoutManager = LinearLayoutManager(view.context)
+        fragmentAgendaBinding.agendaItemRecyclerView.adapter = adapter
+        fragmentAgendaBinding.agendaItemRecyclerView.layoutManager =
+            LinearLayoutManager(view.context)
```
philipplackner, line 68: Instead of `view.context`, we normally use `requireContext()` in Fragments, but both works :)

app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
```diff
@@ -1,64 +1,279 @@
...
+
+        subscribeToObservables()
+        setOnClickListeners()
+    }
+
+    private fun <T> Fragment.collectLatestLifecycleFlow(
```
philipplackner, line 74: I'd put this in a separate file and make it public since you'll likely need this in all of your Fragments. Here you'll never expect such functionality

app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
```diff
@@ -1,64 +1,279 @@
...
+        collectLatestLifecycleFlow(viewModel.dateSelected) { dateSelected ->
+            setupMiniCalendar(dateSelected)
         }
+    }
+
+    private fun setOnClickListeners() {
+
+        //On click listeners for miniCalendar allowing user to quickly change to near future dates
+        fragmentAgendaBinding.apply {
+            today.setOnClickListener {
```
philipplackner, line 99: Might've been better to dynamically occupy this list using a loop. That way you're more flexible when it comes to changing the amount of days you show and it removes quite a bit of boilerplate code here

app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
```diff
@@ -1,64 +1,279 @@
...
+        val todayDate = currentDateAndTime.toLocalDate()
+        val tomorrowDate = currentDateAndTime.toLocalDate().plusDays(1)
+        val thirdDayDate = currentDateAndTime.toLocalDate().plusDays(2)
+        val fourthDayDate = currentDateAndTime.toLocalDate().plusDays(3)
+        val fifthDayDate = currentDateAndTime.toLocalDate().plusDays(4)
+        val sixthDayDate = currentDateAndTime.toLocalDate().plusDays(5)
+
+        //sets up dateNumber text view in the mini calendar
+        ItemMiniCalendarDayBinding.bind(fragmentAgendaBinding.today).dateNumber.text =
+            todayDate.format(DateTimeFormatter.ofPattern("d"))
```
philipplackner, line 209: Instead of re-initializing this `DateTimeFormatter` every time, you could just create it once and use it for all dates

app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
```diff
@@ -1,64 +1,279 @@
...
+            fourthDayDate.format(DateTimeFormatter.ofPattern("eeeee"))
+        ItemMiniCalendarDayBinding.bind(fragmentAgendaBinding.fifthDay).dayOfWeek.text =
+            fifthDayDate.format(DateTimeFormatter.ofPattern("eeeee"))
+        ItemMiniCalendarDayBinding.bind(fragmentAgendaBinding.sixthDay).dayOfWeek.text =
+            sixthDayDate.format(DateTimeFormatter.ofPattern("eeeee"))
+
+        //If dateSelected is on miniCalendar, it sets the background color to yellow
+        if (dateSelected == todayDate) {
+            ItemMiniCalendarDayBinding.bind(fragmentAgendaBinding.today).miniCalendarCard
+                .setCardBackgroundColor(Color.parseColor("#fdeda7"))
```
philipplackner, line 238: I'd avoid using hardcoded colors here. If that requirement changes, it'll become a headache to change them all. Instead, I'd recommend to use color resources (from an XML file)

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -38,16 +41,15 @@ class AgendaItemDetailFragment : Fragment(R.layout.fragment_agenda_item_detail)
...
         agendaItemType = AgendaItemType.valueOf(args.agendaItemType)
 
         subscribeToObservables()
         setupAgendaItemType()
         onClickListeners()
-
     }
 
-    fun <T> Fragment.collectLatestLifecycleFlow(flow: Flow<T>, onCollect: suspend (T) -> Unit) {
+    private fun <T> Fragment.collectLatestLifecycleFlow(flow: Flow<T>, onCollect: suspend (T) -> Unit) {
```
philipplackner, line 52: If you follow my recommendation to move this into its own file like `FragmentExt.kt`, you can remove this one here

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -240,10 +265,76 @@ class AgendaItemDetailFragment : Fragment(R.layout.fragment_agenda_item_detail)
             descriptionTextView.isClickable = isEditing
 
             time.isClickable = isEditing
+            timeButton.isVisible = isEditing
+            timeButton.isClickable=isEditing
             date.isClickable = isEditing
+            dateButton.isVisible = isEditing
+            dateButton.isClickable = isEditing
```
philipplackner, line 272: I think I mentioned this the last time, but this should be `dateButton.isEnabled = isEditing`

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailViewModel.kt
```diff
@@ -39,4 +40,10 @@ class AgendaItemDetailViewModel : ViewModel() {
     fun setSelectedTime(selectedTime: String?){
         _selectedTime.value = selectedTime
     }
+
+    private val _selectedReminderTime = MutableStateFlow("10 minutes before")
```
philipplackner, line 44: Instead of saving the string of the selected item, I'd save the index. If the text changes it's likely you'll forget to change it here which will cause a bug since this text doesn't exist.

app/src/main/res/values/strings.xml
```diff
@@ -24,4 +24,23 @@
     <string name="save_button">Save</string>
     <string name="at">At</string>
     <string name="delete_task_button">DELETE TASK</string>
+
+    <!--Reminder time options popup menu on agenda item detail screen -->
+    <string name="ten_minutes_before">10 minutes before</string>
+    <string name="thirty_minutes_before">30 minutes before</string>
```
philipplackner, line 30: For such strings, you could use placeholder string resources, like this:
`<string name="x_min_before">%d minutes before/>`

And then in your code:
` context.getString(R.string.x_min_before, 10)` will create `10 minutes before`.

That way, you only have one string resource per minute, hour and day.

philipplackner (approved):

PR #12, opened 2022-10-10
Cleans up code

andrewhamilton0:
-Cleans up code
-Adds AgendaItems.kt object and puts agendaItemList and sorting inside
-Makes DeleteConfirmationDialogFragment its own class
-Adds FragmentExt to its own file

### DIFF
--- a/app/src/main/java/com/andrew/tasky/domain/AgendaItems.kt
+++ b/app/src/main/java/com/andrew/tasky/domain/AgendaItems.kt
@@ -1,0 +1,39 @@
+package com.andrew.tasky.domain
+
+import com.andrew.tasky.data.AgendaItem
+import com.andrew.tasky.util.AgendaItemType
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+object AgendaItems {
+
+    private val agendaItemList = mutableListOf(
+        AgendaItem(
+            AgendaItemType.EVENT, false, "Meeting",
+            "Project meeting with Abby", LocalDateTime.of(2022, 10, 8, 8, 22),
+            LocalDateTime.of(2022, 10, 8, 8, 32)
+        ),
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/andrew/tasky/domain/AgendaItems.kt
```diff
@@ -0,0 +1,39 @@
...
+        )
+    )
+
+    //Returns a sorted list of agendaItems by start time and only those of dateSelected
+    fun sortByDateSelected(dateSelected: LocalDate): MutableList<AgendaItem>{
+        val agendaItemListSorted = mutableListOf<AgendaItem>()
+        agendaItemList.sortedBy{ agendaItem -> agendaItem.startDateAndTime }
+            .forEach{if (it.startDateAndTime.toLocalDate() == dateSelected){
+                agendaItemListSorted.add(it)}}
+        return agendaItemListSorted
```
philipplackner, line 37: I'd always work with immutable data types, since they're much less vulnerable to race conditions. Therefore, better return `List<AgendaItem>` here instead.
You rarely need to use a `mutableListOf()` by yourself, usually you're totally fine with the functions from the Kotlin standard lib. So, you could also write this function like this:
```
fun sortByDate(items: List<AgendaItem>, dateSelected: LocalDate): List<AgendaItem> {
    return items
        .filter {
            // Filtering first, so only a smaller list needs to be sorted.
            // If this condition is true, the element will stay in the list, else it won't
            it.startDateAndTime.toLocalDate() == dateSelected 
        } 
        .sortedBy { it.startDateAndTime }
}
```

Also, I'd recommend to pass the list as a parameter to this function, since then you're more flexible and can better reuse this with any list.

philipplackner (approved):

PR #14, opened 2022-10-15
Adds abilitity to delete, add, and replace agenda items

andrewhamilton0:
-Cleans up code
-Allows user to delete, add, and replace agenda items
-Changes colors to be the same as Figma
-Changes AgendaItemDetailFragment xml to be same as Figma
-Adds fonts from Figma
-Adds some icons from Figma
-Adds photo screen (still needs updating to look like figma)
-Adds event exclusive options to agendaItemDetailScreen


app/src/main/java/com/andrew/tasky/domain/AgendaItems.kt
```diff
@@ -52,4 +52,15 @@ object AgendaItems {
...
+
+    fun addAgendaItem(newAgendaItem: AgendaItem){
+        agendaItemList.add(newAgendaItem)
+    }
+
+    fun replaceAgendaItem(newAgendaItem: AgendaItem, oldAgendaItem: AgendaItem){
+        agendaItemList[agendaItemList.indexOf(oldAgendaItem)] = newAgendaItem
+    }
+    fun deleteAgendaItem(agendaItem: AgendaItem){
+        agendaItemList.remove(agendaItem)
```
philipplackner, line 64: Is there a reason why you have all this in a separate class? In the end this list of agenda items is UI state and should be put in a ViewModel. I'd also recommend using an immutable list instead, since mutable ones are very vulnerable to race conditions

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -89,32 +92,56 @@ class AgendaItemDetailFragment: Fragment(R.layout.fragment_agenda_item_detail) {
...
+                viewModel.setTitle(it.title)
+                viewModel.setDescription(it.description)
+                viewModel.setSelectedDate(it.startDateAndTime.toLocalDate())
+                viewModel.setSelectedTime(it.startDateAndTime.toLocalTime())
+                viewModel.setIsDone(it.isDone)
+                viewModel.setSelectedReminderTime(it.reminderTime)
+            }
+            viewModel.setEditMode(args.isInEditMode)
+            viewModel.setInitialSetupToTrue()
+        }
```
philipplackner, line 107: All this is ViewModel specific code and can be put there. To access navigation arguments in your ViewModel, you can do this:
```
class YourViewModel(private val savedStateHandle: SavedStateHandle): ViewModel() { 

    private val agendaItem: AgendaItem? = savedStateHandle.get("agendaItem", null)
 }
```

and your ViewModel then needs to be initialized in your Fragment like this:
`private val viewModel: YourViewModel by viewModels()`

then you can also get rid of this line in your code:
`viewModel = ViewModelProvider(this)[AgendaItemDetailViewModel::class.java]`

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -89,32 +92,56 @@ class AgendaItemDetailFragment: Fragment(R.layout.fragment_agenda_item_detail) {
...
+
     private fun setupAgendaItemType(){
         binding.apply {
             when (agendaItemType) {
                 AgendaItemType.TASK -> {
-                    taskColorBox.setImageResource(R.drawable.task_icon_box)
-                    taskType.text = getString(R.string.task)
+                    taskColorBox.setImageResource(R.drawable.ic_task_box)
+                    agendaItemTypeTextView.text = getString(R.string.task)
+                    addPhotoLayout.visibility = View.GONE
```
philipplackner, line 116: If it's just about toggling visible and gone, you can simply use `addPhotoLayout.isVisible = true/false`

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -89,32 +92,56 @@ class AgendaItemDetailFragment: Fragment(R.layout.fragment_agenda_item_detail) {
...
+                viewModel.setSelectedTime(it.startDateAndTime.toLocalTime())
+                viewModel.setIsDone(it.isDone)
+                viewModel.setSelectedReminderTime(it.reminderTime)
+            }
+            viewModel.setEditMode(args.isInEditMode)
+            viewModel.setInitialSetupToTrue()
+        }
+    }
+
     private fun setupAgendaItemType(){
```
philipplackner, line 110: In general, I'd avoid this approach of having one screen for multiple types of agenda items. It's true that you can avoid a bit of duplicate code like this, but that shouldn't always be the highest priority.

Imagine at some point, requirements change and just the event detail screen will look a lot different than task & reminder screen. That would mean, you'd need to restructure this whole big fragment and create a separate screen for events. This is just not maintainable in the long run.

Instead, I'd create reusable UI components (in separate XML files, for example for the reminder section of an agenda item) which you can then use in all of your separate screens (EventFragment, TaskFragment, ReminderFragment). 
If you're unsure about how to create such reusable UI components, let me know :)

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -334,18 +350,44 @@ class AgendaItemDetailFragment: Fragment(R.layout.fragment_agenda_item_detail) {
...
 
+    private fun getAgendaItemCurrentInfo(): AgendaItem {
+        return AgendaItem(
+            agendaItemType,
+            viewModel.isDone.value,
+            viewModel.title.value,
+            viewModel.description.value,
+            LocalDateTime.of(viewModel.selectedDate.value, viewModel.selectedTime.value),
+            reminderTime = viewModel.selectedReminderTime.value
+        )
```
philipplackner, line 361: That's also a function, you could completely move to the ViewModel, since it only depends on UI state which also sits there.

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -334,18 +350,44 @@ class AgendaItemDetailFragment: Fragment(R.layout.fragment_agenda_item_detail) {
...
+        val newAgendaItem = getAgendaItemCurrentInfo()
+        if(args.agendaItem != null) {
+            args.agendaItem?.let { oldAgendaItem ->
+                AgendaItems.replaceAgendaItem(newAgendaItem,oldAgendaItem)
+            }
+        }
+        else{
+            AgendaItems.addAgendaItem(newAgendaItem)
+        }
+    }
```
philipplackner, line 374: Same for this function. The fragment should only contain logic to observe UI state and change the UI based on that.

PR #15, opened 2022-10-15
Sets up photo recyclerview

andrewhamilton0:


app/src/main/java/com/andrew/tasky/domain/AgendaItem.kt
```diff
@@ -12,5 +13,6 @@ data class AgendaItem(
     val description: String,
     val startDateAndTime: LocalDateTime,
     val endDateAndTime: LocalDateTime? = null,
-    val reminderTime: ReminderTimes
+    val reminderTime: ReminderTimes,
+    val photos: List<Uri>? = null
```
philipplackner, line 17: As I said in WhatsApp, the idea of just putting plain Uris in a list is a bit simplified. What you need here is a way to distinguish between local (represented by a Uri) and remote photos (represented by a URL), since when you're updating an event, you only want to upload the new ones of course.

app/src/main/java/com/andrew/tasky/presentation/adapters/PhotoItemAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.net.Uri
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import com.andrew.tasky.databinding.ItemPhotoAdapterCardBinding
+import com.andrew.tasky.util.AgendaItemDetailFragmentCommunicationWithRV
+
+class PhotoItemAdapter(
+    private var photos: List<Uri>,
+    private val listener: AgendaItemDetailFragmentCommunicationWithRV
```
philipplackner, line 12: Instead of such a long and rather less descriptive name, I'd just use simple Kotlin lambdas. For example:
```
private val openPhoto: (index: Int) -> Unit
```

app/src/main/java/com/andrew/tasky/presentation/adapters/PhotoItemAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+        holder.binding.apply {
+            if (position != photos.size) {
+                image.setImageURI(photos[position])
+                holder.itemView.setOnClickListener {
+                    listener.openPhoto(position)
+                }
+            }
+            if(position != 10){
+                if (position == photos.size){
+                    holder.itemView.setOnClickListener {
```
philipplackner, line 33: You could merge these conditions in one: `if(position != 10 && position == photos.size)`

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -90,8 +104,27 @@ class AgendaItemDetailFragment: Fragment(R.layout.fragment_agenda_item_detail) {
...
+        collectLatestLifecycleFlow(viewModel.photos){
+            val adapter = PhotoItemAdapter(it, this)
+            binding.photosRecyclerView.adapter = adapter
+            binding.photosRecyclerView.layoutManager = LinearLayoutManager(
+                requireContext(), LinearLayoutManager.HORIZONTAL, false)
+            if (it.isEmpty()){
+                binding.addPhotoPlusSign.isVisible = true
+                binding.addPhotoTextView.isVisible = true
+                binding.photosTextView.isVisible = false
+                binding.photosRecyclerView.isVisible = false
```
philipplackner, line 116: Instead of all these true/false values, just use the boolean from the if-statement:
`binding.addPhotoPlusSign.isVisible = it.isEmpty()`

That way, you don't need this if/else

app/src/main/java/com/andrew/tasky/presentation/agenda_item_detail/AgendaItemDetailFragment.kt
```diff
@@ -113,37 +147,60 @@ class AgendaItemDetailFragment: Fragment(R.layout.fragment_agenda_item_detail) {
...
+        addPhotoSearchForResult.launch("image/*")
+    }
+
+    override fun openPhoto(index: Int) {
+        super.openPhoto(index)
+
+        //Listener to Delete Image
+        val supportFragmentManager = requireActivity().supportFragmentManager
+        setFragmentResultListener("REQUEST_KEY1"){
+            resultKey, bundle -> if(resultKey == "REQUEST_KEY1"){
```
philipplackner, line 192: What does `REQUEST_KEY1` represent? Doesn't really communicate anything

philipplackner (approved):

PR #16, opened 2022-10-17
Sets up attendee recyclerview

andrewhamilton0:
-Sets up attendee recyclerview
-adds validation to email, password, and name

### DIFF
--- a/app/src/main/java/com/andrew/tasky/presentation/register/RegisterFragment.kt
+++ b/app/src/main/java/com/andrew/tasky/presentation/register/RegisterFragment.kt
@@ -4,10 +4,13 @@ import androidx.lifecycle.ViewModelProvider
 import android.os.Bundle
 import androidx.fragment.app.Fragment
 import android.view.View
+import androidx.core.view.isVisible
+import androidx.core.widget.addTextChangedListener
 import androidx.navigation.NavController
 import androidx.navigation.Navigation
 import com.andrew.tasky.R
 import com.andrew.tasky.databinding.FragmentRegisterBinding
+import java.util.regex.Pattern
 
 class RegisterFragment : Fragment(R.layout.fragment_register) {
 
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/andrew/tasky/presentation/register/RegisterFragment.kt
```diff
@@ -28,5 +31,14 @@ class RegisterFragment : Fragment(R.layout.fragment_register) {
         binding.backButton.setOnClickListener(){
             navController.popBackStack()
         }
+
+        binding.nameEditText.addTextChangedListener{
+            if (it != null) {
+                binding.nameCheckBox.isVisible = it.length in 2..50
```
philipplackner, line 37: This will not survive config changes. That's UI state that belongs in the ViewModel. For example, a StateFlow called `isNameValid`

app/src/main/java/com/andrew/tasky/presentation/views/PasswordTextField.kt
```diff
@@ -2,20 +2,34 @@ package com.andrew.tasky.presentation.views
...
+    private val passwordPattern = Pattern.compile(
+        passwordRegex
+    )
+
     private val binding = TextFieldPasswordBinding.bind(
         inflate(context, R.layout.text_field_password, this)
     )
 
-    init {
+    fun isPasswordValid(): Boolean{
```
philipplackner, line 27: Personal opinion: Regex helps to implement validation patterns with little code, but it makes your code super hard to read. Nobody will be able to tell you from the pattern what the requirements for a valid password are. Something like this is a lot more readable:
```
class PasswordValidator {

    fun validate(password: String): ValidationResult {
        val containsLowerCase = password.any { it.isLowerCase() }
        val containsUpperCase = password.any { it.isUpperCase() }
        val containsNumber = password.any { it.isDigit() }
        if(password.length < MIN_PASSWORD_LENGTH || !containsLowerCase || !containsUpperCase || !containsNumber) {
            return ValidationResult(
                isSuccessful = false,
                error = UiText.StringResource(R.string.error_invalid_password_format, MIN_PASSWORD_LENGTH)
            )
        }

        return ValidationResult(isSuccessful = true)
    }

    companion object {
        const val MIN_PASSWORD_LENGTH = 9
    }
}
```

Regex is great, but for such easy checks I'd avoid it. It's great to check more complex patterns like an email address.

philipplackner (approved):

PR #18, opened 2022-10-25
Adds delete function for attendee

andrewhamilton0:
-Cleans some code based on recommendations
-Adds attendee data class

### DIFF
--- a/app/src/main/java/com/andrew/tasky/domain/StringToInitials.kt
+++ b/app/src/main/java/com/andrew/tasky/domain/StringToInitials.kt
@@ -1,0 +1,24 @@
+package com.andrew.tasky.domain
+
+class StringToInitials {
+
+    fun convertStringToInitials(name: String):String {
+        // Trims name for crash safety
+        val trimmedName = name.trim()
+        if (trimmedName.isEmpty()) return ""
+
+        //Gets initials and capitalizes it
+        val nameParts = trimmedName.split(' ')
+        return if(nameParts.size > 1){
+            nameParts[0][0].uppercaseChar().toString() +
+                    nameParts[nameParts.lastIndex][0].uppercaseChar().toString()
+        }
+        else {
+            if (nameParts[0].toCharArray().size >= 2){
+                nameParts[0][0].uppercaseChar().toString() +
+                        nameParts[0][1].uppercaseChar().toString()
+            }
+            else nameParts[0].uppercase()
+        }
+    }
+}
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/andrew/tasky/domain/StringToInitials.kt
```diff
@@ -0,0 +1,24 @@
...
+                    nameParts[nameParts.lastIndex][0].uppercaseChar().toString()
+        }
+        else {
+            if (nameParts[0].toCharArray().size >= 2){
+                nameParts[0][0].uppercaseChar().toString() +
+                        nameParts[0][1].uppercaseChar().toString()
+            }
+            else nameParts[0].uppercase()
+        }
+    }
```
philipplackner, line 23: This is an excellent example to practice writing some simple test cases 😁 Have you done that before?

app/src/main/java/com/andrew/tasky/domain/StringToInitials.kt
```diff
@@ -0,0 +1,24 @@
+package com.andrew.tasky.domain
+
+class StringToInitials {
```
philipplackner, line 3: Could be an object as well, since you don't need to pass any parameters to this class

app/src/main/java/com/andrew/tasky/presentation/adapters/AttendeeItemAdapter.kt
```diff
@@ -32,16 +36,15 @@ class AttendeeItemAdapter(
...
-                attendeeInitialsTextView.text = initials
-            }
-            else{
-                val initials = attendees[position].take(2).uppercase()
-                attendeeInitialsTextView.text = initials
+            //Adds initials to card
+            attendeeInitialsTextView.text = StringToInitials().convertStringToInitials(attendees[position].name)
+
+            //Adds on click listener to delete button
+            deleteAttendeeButton.setOnClickListener {
```
philipplackner, line 46: Comments like these just mess up the codebase in the long run. My grandma could probably even tell you what these lines do 😁 
In general, I only comment on something if it's unclear WHY something happens. For example if you add a random `delay(10)` to your code because it might be necessary as a workaround for a bug, that's not clear to anyone. In that case, it totally makes sense to explain in a comment why that delay was added.

If you feel like you need a comment to explain WHAT happens because it's not clear otherwise, that's usually a sign that you can improve your code readability by adding a well named function, variable or class.

philipplackner (approved):

philipplackner:
Here a sample how you can use lambdas in a RecyclerViewAdapter:
```
class MyAdapter(
    // In the first parentheses, you describe the types of parameters your lambda function has
    // After the arrow, you describe the return type of the lambda
    // Here, we want to know the attendee reference that was clicked on
    private val onDeleteIconClick: (Attendee) -> Unit
) {
    // ...

    override fun onBindViewHolder(...) {
        val attendee = attendees[position]
        holder.binding.deleteIcon.setOnClickListener {
            // Here you trigger the lambda function, similar like a function from an interface
            onDeleteIconClick(attendee)
        }
    }
}
```

Then in your Activity/Fragment:
```
val adapter = MyAdapter(
    onDeleteIconClick = { attendee ->
        viewModel.deleteAttendee(attendee)
    }
)
```

PR #19, opened 2022-10-30
-Adds Event Fragment and Restructures it

andrewhamilton0:
-Adds Event Fragment to start seperating agendaItemDetailFragment into three separate fragments,
-Adds a couple new custom views
-Simplified PasswordValidator.kt and NameValidator.kt
-Removed recycler adapter interfaces and used Kotlin Lambdas instead
-Made StringToInitials.kt into an object
-Restructured EventFragment to be more modular


app/src/main/java/com/andrew/tasky/presentation/adapters/AttendeeItemAdapter.kt
```diff
@@ -25,26 +25,25 @@ class AttendeeItemAdapter(
...
             }
-            else{
+            else if (!isAttendee){
                 deleteAttendeeButton.isVisible = true
                 creatorTextView.isVisible = false
             }
+            else{
+                deleteAttendeeButton.isVisible = false
+                creatorTextView.isVisible = false
+            }
```
philipplackner, line 39: Every time, you have an if/else where you just set booleans to true/false, that's very suspicious and you can get rid of the if/else:
```
val isCreator = attendees[position].attendeeType == AttendeeType.CREATOR
deleteAttendeeButton.isVisible = !isCreator
creatorTextView.isVisible = isCreator
```

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailViewModel.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    private val _photos = MutableStateFlow(listOf<Uri>())
+    val photos = _photos.asStateFlow()
+    fun addPhoto(uri: Uri){
+        _photos.value += uri
+    }
+    fun deletePhoto(index1: Int){
+        val updatedPhotos = photos.value.filterIndexed { index, _ -> index != index1  }
+        _photos.value = updatedPhotos
+    }
```
philipplackner, line 81: I'd avoid unclear names like `index1`, that does not communicate the difference to `index` at all. In this case, I'd probably either keep the index in the filter lambda unnamed (and therefore equal to `it`) or name it `currentIndex` or so.

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailViewModel.kt
```diff
@@ -0,0 +1,129 @@
...
+    val selectedShowListOfAttendeesButton = _selectedAttendeeButton.asStateFlow()
+    fun showAllAttendees(){
+        _selectedAttendeeButton.value = ShowListOfAttendeesButtonTypes.ALL
+    }
+    fun showGoingAttendees(){
+        _selectedAttendeeButton.value = ShowListOfAttendeesButtonTypes.GOING
+    }
+    fun showNotGoingAttendees(){
+        _selectedAttendeeButton.value = ShowListOfAttendeesButtonTypes.NOT_GOING
+    }
```
philipplackner, line 116: You could also just create one function for this (BTW, I'd also prefer `AttendeeFilterType` over `ShowListOfAttendeesButtonTypes`):
```
fun setAttendeeFilterType(type: ShowListOfAttendeesButtonTypes) {
    _selectedAttendeeButton.value = type
}
```

app/src/main/java/com/andrew/tasky/presentation/views/TimeDateSelectorCustomView.kt
```diff
@@ -0,0 +1,17 @@
...
+
+class TimeDateSelectorCustomView@JvmOverloads constructor(
+        context: Context,
+        attrs: AttributeSet? = null
+    ): ConstraintLayout(context, attrs) {
+
+        private val binding = CvTimeDateSelectorBinding.bind(
+            inflate(context, R.layout.cv_time_date_selector, this)
+        )
+    }
```
philipplackner, line 17: Is this your whole custom view or will you extend this with some functionality?

philipplackner (approved):

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
```diff
@@ -0,0 +1,723 @@
...
+import java.time.LocalDateTime
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter
+
+class EventDetailFragment : Fragment(R.layout.fragment_event_detail) {
+
+    private val currentDate = LocalDateTime.now()
+        .format(DateTimeFormatter.ofPattern("dd MMMM yyyy")).uppercase()
+
+    private lateinit var viewModel: EventDetailViewModel
```
philipplackner, line 43: I'd prefer this to initialize your VM:
`private val viewModel: EventDetailViewModel by viewModels()`

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
```diff
@@ -0,0 +1,723 @@
...
+                viewModel.setStartDate(it.startDateAndTime.toLocalDate())
+                viewModel.setStartTime(it.startDateAndTime.toLocalTime())
+                viewModel.setIsDone(it.isDone)
+                viewModel.setSelectedReminderTime(it.reminderTime)
+                it.photos?.let { photoList -> viewModel.setupPhotos(photoList) }
+            }
+            viewModel.setEditMode(args.isInEditMode)
+            viewModel.setInitialSetupToTrue()
+        }
+    }
```
philipplackner, line 111: Normally, this is stuff shouldn't need to initialize in the fragment since it's all ViewModel code. You can retrieve navigation arguments in a ViewModel like this:
```
class EventDetailViewModel(
    private val savedStateHandle: SavedStateHandle
): ViewModel() {
    
    init {
        savedStateHandle.get<AgendaItem>("agendaItem")?.let { item -> 
            // Update all the states here
        }
    }
}
```

Normally, the savedStateHandle instance should be passed automatically to your VM when you construct it using `by viewModels()` (not 100% on this anymore, if this doesn't work for you let me know)

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
```diff
@@ -0,0 +1,723 @@
...
+                EventDetailFragmentDirections
+                    .actionEventDetailFragmentToPhotoDetailFragment(
+                        viewModel.photos.value[index].toString(), index
+                    )
+            )
+        }
+
+        binding.apply {
+            collectLatestLifecycleFlow(viewModel.photos) { photoList ->
+                addPhotoLayout.isVisible = !(isAttendee && photoList.isEmpty())
```
philipplackner, line 232: If you negate a whole boolean condition, it's equivalent to negating all single parts and swapping && to || or vice versa:
`!(isAttendee && photoList.isEmpty()) == !isAttendee || photoList.isNotEmpty()`

The latter is a bit easier to process in your brain IMO

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
```diff
@@ -0,0 +1,723 @@
...
+
+            if (!isAttendee) {
+                collectLatestLifecycleFlow(viewModel.isInEditMode) { isEditing ->
+                    timeTextView.isEnabled = isEditing
+                    timeButton.isEnabled = isEditing
+                    timeButton.isVisible = isEditing
+                    dateTextView.isEnabled = isEditing
+                    dateButton.isEnabled = isEditing
+                    dateButton.isVisible = isEditing
+                }
```
philipplackner, line 284: I'd personally prefer not structuring setting up your UI based on the sections of your layout, but rather based on the states you observe. Here, you'll add multiple observers to `isInEditMode`, since you call `collectLatestLifecycleFlow` at multiple places. One observer would be enough though. I don't really see a reason at all why you need to put setting up every section into its own function, I'd just have one `setupObservers()` or so in which you add all your flow collectors. That still makes it easy to find stuff.

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
```diff
@@ -0,0 +1,723 @@
...
```
philipplackner, line 334: And here's the 3rd observer for this 😁

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
```diff
@@ -0,0 +1,723 @@
...
+        }
+        collectLatestLifecycleFlow(viewModel.notGoingAttendees){ notGoingAttendees ->
+            val notGoingAttendeeAdapter = AttendeeItemAdapter(
+                notGoingAttendees,
+                isAttendee,
+                onDeleteIconClick = {attendee ->  viewModel.deleteAttendee(attendee)})
+            binding.notGoingAttendeeRecyclerView.adapter = notGoingAttendeeAdapter
+            binding.notGoingAttendeeRecyclerView.layoutManager = LinearLayoutManager(requireContext())
+        }
+    }
```
philipplackner, line 526: I wouldn't recreate the `LinearLayoutManager` on every emission here. You just need to setup your RecyclerView once, but then you don't need to assign adapter and layout manager again. Just toggle its visibility and you're good.
So all that should happen here is that you update the attendees list of your adapter. Same counts for the `goingAttendees` collector above.

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
```diff
@@ -0,0 +1,723 @@
...
+                resultKey, bundle -> if(resultKey == "REQUEST_KEY"){
+            val deleteAgendaItem = bundle.getBoolean("DELETE_AGENDA_ITEM")
+            if (deleteAgendaItem){
+                args.agendaItem?.let {
+                    AgendaItems.deleteAgendaItem(it)
+                }
+                navController.popBackStack()
+            }
+        }
+        }
```
philipplackner, line 715: Here, the formatting looks quite messed up, I'd recommend taking a look at setting up a linting & auto-formatting tool like ktlint to enforce a consistent style. Here I explain it:
https://www.youtube.com/watch?v=hSgPNbEcX98

philipplackner (approved):

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
andrewhamilton0: So I should just move "binding.notGoingAttendeeRecyclerView.layoutManager = LinearLayoutManager(requireContext())" outside the scope of the flow? And the same as goingAttendees

app/src/main/java/com/andrew/tasky/presentation/views/TimeDateSelectorCustomView.kt
andrewhamilton0: I think this is the only functionality I was gonna use for this

philipplackner: Then you don't need a custom view for this, you can instead use this in your XML layout to get a reference to the corresponding binding:
```
<include android:id="@+id/timeDateSelectorView" layout="@layout/your_layout_here" />
```

app/src/main/java/com/andrew/tasky/presentation/event_detail/EventDetailFragment.kt
philipplackner: Yes, the only thing you need to update is the list the RV holds

PR #20, opened 2022-11-04
-Added add attendee dialog

andrewhamilton0:
-Added add attendee dialog
-Added KTLINT plugin
-Updated miniCalendar on agenda screen to use a recyclerView
-Updated edit screen to look more like mockup
-Updated agenda screen to look more like mockup
-Added save state handle to event fragment
-Put code into fragment_reminder_detail.xml and fragment_task_detail.xml
-Added unit test for StringToInitials.kt (StringToInitialsTest.kt)
-AGENDA NAVIGATIONS FROM AGENDA SCREEN DO NOT WORK CORRECTLY AS I AM STILL MIGRATING FROM ONE AGENDA DETAIL SCREEN TO THREE SEPARATE ONES


app/src/main/java/com/andrew/tasky/domain/AgendaItems.kt
```diff
@@ -44,28 +44,31 @@ object AgendaItems {
...
-        agendaItemList.sortedBy{ agendaItem -> agendaItem.startDateAndTime }
-            .forEach{if (it.startDateAndTime.toLocalDate() == dateSelected){
-                agendaItemListSorted.add(it)}}
+        agendaItemList.sortedBy { agendaItem -> agendaItem.startDateAndTime }
+            .forEach {
+                if (it.startDateAndTime.toLocalDate() == dateSelected) {
+                    agendaItemListSorted.add(it)
+                }
+            }
         return agendaItemListSorted
```
philipplackner, line 56: Better directly avoid using mutable lists because race conditions can easily happen:
```
return agendaItemList.sortedBy { agendaItem ->
    agendaItem.startDateAndTime
}.filter {
    it.startDateAndTime.toLocalDate() == dateSelected
}
```

app/src/main/java/com/andrew/tasky/presentation/adapters/MiniCalendarAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+
+            if (dateSelected == startDate.plusDays(position.toLong())) {
+                miniCalendarCard.setCardBackgroundColor(
+                    ResourcesCompat.getColor(
+                        holder.itemView.context.resources,
+                        R.color.orange,
+                        null
+                    )
+                )
+            } else {
```
philipplackner, line 45: This adapter would be a bit more flexible if you'd instead pass in a `List<LocalDate>`. In case you'd like to add some scrolling + pagination in future (in both directions), the current approach wouldn't work since you assume there is a start date.
Doesn't need to be fixed, just mentioning :)

app/src/main/java/com/andrew/tasky/presentation/edit/EditFragment.kt
```diff
@@ -53,9 +53,11 @@ class EditFragment : Fragment(R.layout.fragment_edit) {
                     when (it) {
                         EditType.DESCRIPTION -> {
                             binding.editTypeTitle.text = getString(R.string.edit_description_text)
+                            binding.inputEditText.textSize = 16F
                         }
                         EditType.TITLE -> {
                             binding.editTypeTitle.text = getString(R.string.edit_title_text)
+                            binding.inputEditText.textSize = 26F
```
philipplackner, line 60: Don't hardcode your text sizes in pixels, use SP units. These scale together with the user's font size preference. Android offers functions to convert these to pixels.

philipplackner (approved):

PR #26, opened 2022-11-15
Added ReminderDetailFragment code

andrewhamilton0:
-Added ReminderDetailFragment.kt code from EventDetailFragment.kt 
-Added Photo data class
-Nullable items now work in Converters.kt

### DIFF
--- a/app/src/main/java/com/andrew/tasky/presentation/reminder_detail/ReminderDetailFragment.kt
+++ b/app/src/main/java/com/andrew/tasky/presentation/reminder_detail/ReminderDetailFragment.kt
@@ -1,32 +1,341 @@
 package com.andrew.tasky.presentation.reminder_detail
 
+import android.graphics.Paint
 import android.os.Bundle
-import android.view.LayoutInflater
-import android.view.View
-import android.view.ViewGroup
+import android.view.*
+import android.widget.PopupMenu
+import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/andrew/tasky/presentation/reminder_detail/ReminderDetailFragment.kt
```diff
@@ -1,32 +1,341 @@
...
+                bundle.putString("EDIT_TYPE", editType.name)
+
+                setFragmentResult("EDIT_TYPE_AND_TEXT_REQUEST_KEY", bundle)
+
+                navController.navigate(
+                    ReminderDetailFragmentDirections
+                        .actionReminderDetailFragmentToEditFragment()
+                )
+            }
+        }
```
philipplackner, line 238: You could think about making this an extension function of `Fragment`, since you'll otherwise need to have this same code in all 3 types of fragments. The function signature could look something like this:
```
fun Fragment.navigateToEditFragment(editType: EditType, onResult: (String) -> Unit) {
    val navController = findNavController()
    // Put the same code in here
}
```

Call it like this, for example in `ReminderDetailFragment`:
```
navigateToEditFragment(
    editType = EditType.TITLE,
    onResult = viewModel::setTitle
)
```

PS: You'll also need to make the navigation to the edit fragment a global navigation, so you can call it from any fragment and not only those where you set a direct transition to it.

philipplackner (approved):

PR #27, opened 2022-11-17
Changes adapters to use ListAdapter DiffUtil

andrewhamilton0:


app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaFragment.kt
```diff
@@ -41,22 +42,21 @@ class AgendaFragment : Fragment(R.layout.fragment_agenda) {
         setOnClickListeners()
         setupCurrentMonthTextView()
         setupAgendaItemListRecyclerView()
-    }
-
-    private fun subscribeToObservables() {
+        miniCalendarAdapter = MiniCalendarAdapter(
+            onHolderClick = { date -> viewModel.setDateSelected(date) }
```
philipplackner, line 46: Would rather call this `onDateClick` since from the outside it's not clear what the `holder` is

app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaViewModel.kt
```diff
@@ -31,11 +32,47 @@ class AgendaViewModel@Inject constructor(
 
     fun setDateSelected(dateUserSelected: LocalDate) {
         _dateSelected.value = dateUserSelected
+        _calendarDateItemList.value.map { item ->
+            item.isSelected = dateUserSelected == item.date
+        }
```
philipplackner, line 37: Here, you're not changing the actual state since you don't do anything with the result. To assign a new state, you have to set it to `_calendarDateItemList` like this:
```
_calendarDateItemList.value = _calendarDateItemList.value.map { ... }
```

This causes the state to not be changed which therefore doesn't trigger your collector in the UI and therefore doesn't properly update your background in the date list.

app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaViewModel.kt
```diff
@@ -31,11 +32,47 @@ class AgendaViewModel@Inject constructor(
...
+            )
+        }
+        for (day in 1..daysAfter) {
+            _calendarDateItemList.value += CalendarDateItem(
+                false,
+                LocalDate.now().plusDays(
+                    day.toLong()
+                )
+            )
+        }
```
philipplackner, line 73: I'd first calculate the final list you want to assign and then assign it, since this way, you'll trigger many UI updates (with every call on `+=`)

philipplackner (approved):

PR #31, opened 2022-12-15
Combines date selected and time flow for mini calendar

andrewhamilton0:
-Combines date selected and time flow for mini calendar
-Moves agenda item menu options from agenda adapter to agenda fragment 
-Makes fragment_agenda.xml scrollable

### DIFF
--- a/app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaViewModel.kt
+++ b/app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaViewModel.kt
@@ -21,8 +21,6 @@ class AgendaViewModel@Inject constructor(
 
     private val _dateSelected = MutableStateFlow(LocalDate.now())
     val dateSelected = _dateSelected.asStateFlow()
-    private var _currentDateAndTime = MutableStateFlow(LocalDateTime.now())
-    private val currentDateAndTime = _currentDateAndTime.asStateFlow()
 
     private val currentDateAndTimeFlow = flow<LocalDateTime> {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/andrew/tasky/presentation/agenda/AgendaViewModel.kt
```diff
@@ -72,43 +58,39 @@ class AgendaViewModel@Inject constructor(
...
+
+    private val daysAfterCurrentDate = 5
+    val calendarDateItemList = combine(
+        dateSelected, currentDateAndTimeFlow
+    ) { dateSelected, currentDateTime ->
+        (0..daysAfterCurrentDate).map { days ->
             CalendarDateItem(
-                isSelected = dateUserSelected == item.date,
-                date = item.date
+                isSelected = dateSelected == currentDateTime.toLocalDate().plusDays(days.toLong()),
```
philipplackner, line 85: Maybe I'm getting this wrong, but why do you call `plusDays(days.toLong())` here?
Shouldn't the selected date simply be the currentDate?
`isSelected = dateSelected == currentDateTime.toLocalDate()`

andrewhamilton0: If I use only currentDateTime.toLocalDate() it only compares the date I select on the mini calendar to the date it currently is (LocalDate.now()). So if I removed plusDays(days.toLong()) it'll highlight all days if the dateSelected is today's date and not highlight any if the dateSelected is any other day

PR #33, opened 2022-12-22
Starting to build api connectivity for authentication

andrewhamilton0:
-Starting to build api connectivity for authentication
-Made multiple date types interface


app/src/main/java/com/andrew/tasky/agenda/presentation/screens/agenda/AgendaViewModel.kt
```diff
@@ -88,6 +89,26 @@ class AgendaViewModel@Inject constructor(
...
+                DateType.Today
+            }
+            currentDate.plusDays(1) -> {
+                DateType.Tomorrow
+            }
+            else -> {
+                DateType.FullDate(dateSelected)
+            }
+        }
+    }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), DateType.Today)
```
philipplackner, line 110: Good job with this 👍

app/src/main/java/com/andrew/tasky/auth/presentation/screens/login/LoginFragment.kt
```diff
@@ -2,30 +2,65 @@ package com.andrew.tasky.auth.presentation.screens.login
...
+            viewModel.authResults.collect { result ->
+                when (result) {
+                    is AuthResult.Authorized -> {
+                        navController.navigate(
+                            LoginFragmentDirections.actionLoginFragmentToAgendaFragment()
+                        )
+                        navController.popBackStack(
+                            destinationId = R.id.action_loginFragment_to_agendaFragment,
+                            inclusive = true
+                        )
```
philipplackner, line 50: You can also combine these in one call:
```
navController.navigate(
    LoginFragmentDirections.actionLoginFragmentToAgendaFragment(),
    NavOptions.Builder().setPopUpTo(R.id.agendaFragment, inclusive = true)
)
```

app/src/main/java/com/andrew/tasky/auth/presentation/screens/register/RegisterFragment.kt
```diff
@@ -2,45 +2,87 @@ package com.andrew.tasky.auth.presentation.screens.register
...
             }
-        }
 
-        binding.getStartedButton.setOnClickListener {
-            binding.passwordTextField.isPasswordValid()
+            getStartedButton.setOnClickListener {
+                if (
+                    passwordTextField.isPasswordValid() &&
+                    viewModel.isNameValid.value &&
+                    emailTextField.isValid()
```
philipplackner, line 47: Would it be possible to move this condition in the ViewModel in `register()`?
I like to keep the UI as dumb as possible about such logic, so it only needs to display UI and send actions to the VM. That's helpful because you can easier test logic of a ViewModel than logic in the UI (here you would need to run an actual UI test that simulates clicks etc)

philipplackner (approved):

PR #34, opened 2022-12-31
Adds EventPhoto Sealed Interface

andrewhamilton0:
-Adds EventPhoto Sealed Interface
-Adds Glide Api

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/AttendeeItemAdapter.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/AttendeeItemAdapter.kt
@@ -42,10 +42,10 @@ class AttendeeItemAdapter(
             deleteAttendeeButton.isVisible = !isCreatorHolder && !isAttendee
             creatorTextView.isVisible = isCreatorHolder
 
-            attendeeFullNameTextView.text = currentList[position].name
+            attendeeFullNameTextView.text = currentList[position].fullName
 
             attendeeInitialsTextView.text = StringToInitials
-                .convertStringToInitials(currentList[position].name)
+                .convertStringToInitials(currentList[position].fullName)
 
             deleteAttendeeButton.setOnClickListener {
                 onDeleteIconClick(currentList[position])
--- a/app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
@@ -1,26 +1,31 @@
 package com.andrew.tasky.agenda.presentation.adapters
 
+import android.content.Context
 import android.net.Uri
 import android.view.LayoutInflater
 import android.view.ViewGroup
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ListAdapter
 import androidx.recyclerview.widget.RecyclerView
+import com.andrew.tasky.agenda.domain.models.EventPhoto
 import com.andrew.tasky.agenda.domain.models.Photo
+import com.andrew.tasky.agenda.presentation.screens.event_detail.EventDetailFragment
 import com.andrew.tasky.databinding.ItemPhotoAdapterCardBinding
+import com.bumptech.glide.Glide
 
 class PhotoItemAdapter(
+    val context: Context,
     private val onPhotoClick: (Int) -> Unit,
     private val onAddPhotoClick: () -> Unit,
     private val userIsAttendee: Boolean
-) : ListAdapter<Photo, PhotoItemAdapter.PhotoItemViewHolder>(Companion) {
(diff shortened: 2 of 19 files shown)


app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/AttendeeItemAdapter.kt
```diff
@@ -42,10 +42,10 @@ class AttendeeItemAdapter(
             deleteAttendeeButton.isVisible = !isCreatorHolder && !isAttendee
             creatorTextView.isVisible = isCreatorHolder
 
-            attendeeFullNameTextView.text = currentList[position].name
+            attendeeFullNameTextView.text = currentList[position].fullName
```
philipplackner, line 45: You need this all the time, so better store the result of `currentList[position]` in a variable (and avoid weird bugs like we had before 😁 )

app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
```diff
@@ -1,26 +1,31 @@
...
+    companion object : DiffUtil.ItemCallback<EventPhoto>() {
+        override fun areItemsTheSame(oldItem: EventPhoto, newItem: EventPhoto): Boolean {
             return oldItem == newItem
         }
 
-        override fun areContentsTheSame(oldItem: Photo, newItem: Photo): Boolean {
+        override fun areContentsTheSame(oldItem: EventPhoto, newItem: EventPhoto): Boolean {
             return oldItem == newItem
         }
     }
```
philipplackner, line 31: Careful here, since your `EventPhoto.Local` is not a data class, therefore it does not have an actual equals function to compare its content. So `localPhoto1 == localPhoto2` will only return true if it's the same reference, but not necessarily if it's the same Uri (which you also wouldn't want since a user could add the same photo twice).

I'd recommend also using keys for local photos which you can randomly generate when you add one (using `UUID.randomUUID().toString()`). That way you can better compare if 2 local photos are really the same or not.

app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
```diff
@@ -36,15 +41,21 @@ class PhotoItemAdapter(
 
     override fun onBindViewHolder(holder: PhotoItemViewHolder, position: Int) {
         holder.binding.apply {
+
             if (position != currentList.size) {
-                image.setImageURI(Uri.parse(currentList[position].uriString))
-                holder.itemView.setOnClickListener {
-                    onPhotoClick(position)
+                if (currentList[position] is EventPhoto.Local) {
```
philipplackner, line 46: Same comment as above regarding currentList

app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
```diff
@@ -36,15 +41,21 @@ class PhotoItemAdapter(
...
+                    val item = currentList[position] as EventPhoto.Local
+                    image.setImageURI(Uri.parse(item.uri.toString()))
+                    holder.itemView.setOnClickListener {
+                        onPhotoClick(position)
+                    }
+                } else {
+                    val item = currentList[position] as EventPhoto.Remote
```
philipplackner, line 53: I think I commented before that I'd use if/else if you only have 2 branches.. This one here is actually an exception where I'd prefer a when expression, since you then don't need to cast the result. 
Imagine, at some point you add a 3rd type of photos to `EventPhoto`, then this piece of code might break, while a when-expression will force you to also handle the new case.

app/src/main/java/com/andrew/tasky/auth/presentation/screens/register/RegisterViewModel.kt
```diff
@@ -22,18 +25,28 @@ class RegisterViewModel @Inject constructor(
...
-            val result = repository.register(
-                name = name,
-                email = email,
-                password = password
-            )
-            resultChannel.send(result)
+        if (
+            isPasswordValid(password) &&
+            isNameValid.value &&
+            Patterns.EMAIL_ADDRESS.matcher(email).matches()
```
philipplackner, line 40: `Patterns` is a class from the Android SDK, which means this ViewModel won't be locally unit testable. To avoid this, you can do this:
*In domain*
```
interface EmailPatternValidator {
    fun isValidEmailPattern(email: String): Boolean
}
```

*In data*
```
class EmailPatternValidatorImpl: EmailPatternValidator {

    override fun isValidEmailPattern(email: String): Boolean {
        return Patterns.EMAIL_ADDRESS.matcher(email).matches()
    }
}
```

Then inject an instance of `EmailPatternValidator` in your ViewModel, so it's not forced to use your specific implementation. In local unit tests you could then provide a fake version of it that doesn't use the underlying Android-dependency.

philipplackner (approved):

PR #35, opened 2023-01-05
Adds Agenda Api Interface

andrewhamilton0:
-Adds Agenda Api Interface
-Moved A few files in auth package to a new 'data' folder within auth package
-Added EmailPatternValidatorImpl

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/AgendaApi.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/AgendaApi.kt
@@ -1,0 +1,53 @@
+package com.andrew.tasky.agenda.data
+
+import com.andrew.tasky.agenda.util.Event
+import java.util.TimeZone
+import okhttp3.MultipartBody
+import retrofit2.Response
+import retrofit2.http.DELETE
+import retrofit2.http.GET
+import retrofit2.http.Multipart
+import retrofit2.http.POST
+import retrofit2.http.PUT
+import retrofit2.http.Part
+import retrofit2.http.Query
+
+interface AgendaApi {
+
+    @GET("/agenda")
+    suspend fun getAgendaItems(
+        @Query("timezone")
+        timeZone: TimeZone,
+        @Query("time")
+        time: Long = System.currentTimeMillis()
+    ): Response<AgendaResponse>
+
+    @POST("/syncAgenda")
+    suspend fun syncAgendaItems(): Response<SyncAgendaResponse>
+
+    @Multipart
+    @POST("/event")
+    suspend fun createEvent(
+        @Part eventData: MultipartBody.Part,
+        @Part photoData: MultipartBody.Part
+    ): Response<Event>
+
+    @GET("/event")
+    suspend fun getEvent(
+        @Query("eventId")
+        eventId: String
+    ): Response<Event>
+
+    @DELETE("/event")
+    suspend fun deleteEvent(
+        @Query("eventId")
+        eventId: String
+    )
+
+    @Multipart
+    @PUT("/event")
+    suspend fun updateEvent(
+        @Part eventData: MultipartBody.Part,
+        @Part photoData: MultipartBody.Part
+    ): Response<Event>
+}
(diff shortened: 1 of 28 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/AgendaApi.kt
```diff
@@ -0,0 +1,53 @@
...
+import retrofit2.http.PUT
+import retrofit2.http.Part
+import retrofit2.http.Query
+
+interface AgendaApi {
+
+    @GET("/agenda")
+    suspend fun getAgendaItems(
+        @Query("timezone")
+        timeZone: TimeZone,
```
philipplackner, line 20: Have you tried if this works? Will it automatically serialize the `TimeZone` class to a string?

app/src/main/java/com/andrew/tasky/agenda/data/AgendaApi.kt
```diff
@@ -0,0 +1,53 @@
...
+    ): Response<AgendaResponse>
+
+    @POST("/syncAgenda")
+    suspend fun syncAgendaItems(): Response<SyncAgendaResponse>
+
+    @Multipart
+    @POST("/event")
+    suspend fun createEvent(
+        @Part eventData: MultipartBody.Part,
+        @Part photoData: MultipartBody.Part
```
philipplackner, line 32: This should be a `List<MultipartBody.Part>` since you can attach multiple photos

app/src/main/java/com/andrew/tasky/agenda/data/AgendaApi.kt
```diff
@@ -0,0 +1,53 @@
...
+    suspend fun deleteEvent(
+        @Query("eventId")
+        eventId: String
+    )
+
+    @Multipart
+    @PUT("/event")
+    suspend fun updateEvent(
+        @Part eventData: MultipartBody.Part,
+        @Part photoData: MultipartBody.Part
```
philipplackner, line 51: Same here

app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/AttendeeItemAdapter.kt
```diff
@@ -38,17 +38,18 @@ class AttendeeItemAdapter(
     override fun onBindViewHolder(holder: AttendeeItemViewHolder, position: Int) {
         holder.binding.apply {
 
-            val isCreatorHolder = currentList[position].attendeeType == AttendeeType.CREATOR
+            val item = currentList[position]
+
+            val isCreatorHolder = item.attendeeType == AttendeeType.CREATOR
```
philipplackner, line 43: Here, I'd prefer using a simple boolean for this field (inside the attendee class). The reason is that attendee types might not be mutually exclusive, which an enum class suggests though. So imagine every attendee would also have a certain role or so, maybe `Organizer` or `Presenter`. Your current implementation doesn't allow you to combine these, so that an attendee could be both organizer and creator. 
Right now it works fine as it is because you only use this for visitors and creators, but in case you'd extend this a boolean `isCreator` would make things easier

app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
```diff
@@ -22,11 +20,19 @@ class PhotoItemAdapter(
 
     companion object : DiffUtil.ItemCallback<EventPhoto>() {
         override fun areItemsTheSame(oldItem: EventPhoto, newItem: EventPhoto): Boolean {
-            return oldItem == newItem
+            return if (oldItem is EventPhoto.Local && newItem is EventPhoto.Local) {
+                oldItem.id == newItem.id
+            } else if (oldItem is EventPhoto.Remote && newItem is EventPhoto.Remote) {
+                oldItem.key == newItem.key
+            } else false
```
philipplackner, line 27: Here you would also share the identifier and just have one key attribute for both (in the constructor of the `EventPhoto` sealed class)

philipplackner (approved):

PR #36, opened 2023-01-09
Adds ApiKeyInterceptor

andrewhamilton0:
-Adds API KEY ACCESS through build.gradle defaultConfig and local.properties
-Adds ApiKeyInterceptor

### DIFF
--- a/app/src/main/java/com/andrew/tasky/core/data/RetrofitInstance.kt
+++ b/app/src/main/java/com/andrew/tasky/core/data/RetrofitInstance.kt
@@ -1,0 +1,25 @@
+package com.andrew.tasky.core.data
+
+import com.andrew.tasky.agenda.data.AgendaApi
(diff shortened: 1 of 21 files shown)


app/src/main/java/com/andrew/tasky/core/data/RetrofitInstance.kt
```diff
@@ -0,0 +1,25 @@
+package com.andrew.tasky.core.data
+
+import com.andrew.tasky.agenda.data.AgendaApi
+import com.andrew.tasky.auth.data.AuthApi
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+
+object RetrofitInstance {
```
philipplackner, line 8: Any reason you use Dagger-Hilt but decided to make this a non-dagger singleton?

philipplackner (approved):

PR #39, opened 2023-01-11
1/11/2023 1 adds agenda api rep

andrewhamilton0:


app/src/main/java/com/andrew/tasky/agenda/data/AgendaApiRepositoryImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+            AuthResult.Authorized(taskAsAgendaItem)
+        } catch (e: HttpException) {
+            if (e.code() == 401) {
+                AuthResult.Unauthorized()
+            } else {
+                AuthResult.UnknownError()
+            }
+        } catch (e: Exception) {
+            AuthResult.UnknownError()
+        }
```
philipplackner, line 54: This try/catch logic is logic you'll need in almost every API call, could it make sense to create a helper function to shorten this?

app/src/main/java/com/andrew/tasky/agenda/data/ReminderTimeConversion.kt
```diff
@@ -0,0 +1,47 @@
...
+            ReminderTime.ONE_DAY_BEFORE
+        } else if (duration.seconds.minutes.inWholeHours.toInt() == 6) {
+            ReminderTime.SIX_HOURS_BEFORE
+        } else if (duration.seconds.minutes.inWholeHours.toInt() == 1) {
+            ReminderTime.ONE_HOUR_BEFORE
+        } else if (duration.seconds.minutes.inWholeMinutes.toInt() == 30) {
+            ReminderTime.THIRTY_MINUTES_BEFORE
+        } else {
+            ReminderTime.TEN_MINUTES_BEFORE
+        }
```
philipplackner, line 29: I usually find when-expressions more readable than `else if`, but personal preference

app/src/main/java/com/andrew/tasky/agenda/data/ReminderTimeConversion.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    fun toEpochSecond(startTimeEpochSecond: Long, reminderTime: ReminderTime): Long {
+        val startTime = LocalDateTime.ofEpochSecond(startTimeEpochSecond, 0, null)
+        return when (reminderTime) {
+            ReminderTime.TEN_MINUTES_BEFORE ->
+                startTime.minusMinutes(10).toEpochSecond(null)
```
philipplackner, line 36: What's `null` here?

app/src/main/java/com/andrew/tasky/agenda/data/Task.kt
```diff
@@ -1,4 +1,4 @@
-package com.andrew.tasky.agenda.util
+package com.andrew.tasky.agenda.data
```
philipplackner, line 1: `Task` sounds more like a domain model's name to me. In the data layer I would expect `TaskDto` or `TaskEntity` 🤔

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/reminder_detail/ReminderDetailViewModel.kt
```diff
@@ -68,7 +69,9 @@ class ReminderDetailViewModel @Inject constructor(
 
     fun saveAgendaItem() {
         val agendaItem = AgendaItem(
-            savedStateHandle.get<AgendaItem>("agendaItem")?.id,
+            id = savedStateHandle.get<AgendaItem>("agendaItem")?.id,
+            apiId = savedStateHandle.get<AgendaItem>("agendaItem")?.apiId
+                ?: UUID.randomUUID().toString(),
```
philipplackner, line 74: What's the difference between these two?

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/task_detail/TaskDetailViewModel.kt
```diff
@@ -21,7 +23,8 @@ import kotlinx.coroutines.withContext
 @HiltViewModel
 class TaskDetailViewModel @Inject constructor(
     private val savedStateHandle: SavedStateHandle,
-    private val repository: AgendaItemRepository
+    private val localRepository: AgendaItemRepository,
+    private val remoteRepository: AgendaApiRepository
```
philipplackner, line 27: The purpose of a repository is to combine multiple data sources, so the caching would be part of it, which is why you wouldn't need to distinguish between local and remote repo

app/src/main/java/com/andrew/tasky/core/di/AppModule.kt
```diff
@@ -28,15 +33,29 @@ object AppModule {
...
+            .create()
+    }
+
+    @Provides
+    @Singleton
+    fun provideAgendaApi(prefs: SharedPreferences): AgendaApi {
+        return Retrofit.Builder()
+            .baseUrl("https://tasky.pl-coding.com")
+            .client(
+                OkHttpClient.Builder()
```
philipplackner, line 53: I'd create a separate provides function for this client since you need it for all your API interfaces. If you want to configure it, you would need to change it at all spots instead of just one

app/src/main/java/com/andrew/tasky/agenda/data/ReminderTimeConversion.kt
andrewhamilton0: Offset. It's a java function so I was unable to name it

PR #40, opened 2023-01-11
-Adds more to agendaApiRepo and AgendaApiRepoIml

andrewhamilton0:


app/src/main/java/com/andrew/tasky/agenda/data/AgendaApiRepository.kt
```diff
@@ -1,13 +1,24 @@
...
+    suspend fun deleteAttendee(eventId: String): AuthResult<Unit>
     suspend fun createTask(task: AgendaItem): AuthResult<Unit>
     suspend fun updateTask(task: AgendaItem): AuthResult<Unit>
     suspend fun getTask(taskId: String): AuthResult<AgendaItem>
     suspend fun deleteTask(taskId: String): AuthResult<Unit>
+    suspend fun createReminder(reminderResponse: Reminder): AuthResult<Unit>
+    suspend fun updateReminder(reminderResponse: Reminder): AuthResult<Unit>
+    suspend fun getReminder(reminderId: String): AuthResult<AgendaItem>
+    suspend fun deleteReminder(reminderId: String): AuthResult<Unit>
 }
```
philipplackner, line 24: As you can see this quickly becomes quite crowded, I'd separate repos based on type of model they deal with, so for example `EventRepository`, `TaskRepository` and so on.

app/src/main/java/com/andrew/tasky/agenda/data/AgendaApiRepositoryImpl.kt
```diff
@@ -10,23 +10,165 @@ import retrofit2.HttpException
...
+                    isDone = task.isDone,
+                    title = task.title,
+                    description = task.description ?: "",
+                    startDateAndTime = LocalDateTime.ofEpochSecond(task.time, 0, null),
+                    reminderTime = ReminderTimeConversion.toEnum(
+                        startTimeEpochSecond = task.time,
+                        remindAtEpochSecond = task.remindAt
+                    )
+                )
+            }
```
philipplackner, line 32: This mapping logic is needed very often which is why we usually create mapper functions, like this:
```
fun TaskDto.toTask(): Task {
    return AgendaItem(...)
}
```

app/src/main/java/com/andrew/tasky/agenda/data/AgendaApiRepositoryImpl.kt
```diff
@@ -10,23 +10,165 @@ import retrofit2.HttpException
...
+        timezone: String,
+        time: Long
     ): AuthResult<List<AgendaItem>> {
-        TODO("Not yet implemented")
+        return try {
+            val response = api.getAgendaItems(timezone = timezone, time = time)
+            val tasks = response.tasks.map { task ->
+                AgendaItem(
+                    apiId = task.id,
+                    type = AgendaItemType.TASK,
```
philipplackner, line 22: Didn't you use a sealed class for agenda items? Something like this:
```
sealed class AgendaItem(...) {
    data class Event(...): AgendaItem(...)
    data class Task(...): AgendaItem(...)
    data class Reminder(...): AgendaItem(...)
}
```

That way you don't need the type parameter since it's contained in the class already

PR #41, opened 2023-01-12
1/12/2023 adds UI event photo

andrewhamilton0:
Adds UiEventPhoto to add add photo item
Gets isAttendee from viewModel for EventDetail

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
@@ -1,13 +1,13 @@
 package com.andrew.tasky.agenda.presentation.adapters
 
 import android.content.Context
+import android.net.Uri
 import android.view.LayoutInflater
 import android.view.ViewGroup
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ListAdapter
 import androidx.recyclerview.widget.RecyclerView
 import com.andrew.tasky.agenda.domain.models.EventPhoto
-import com.andrew.tasky.agenda.util.UiEventPhoto
 import com.andrew.tasky.databinding.ItemPhotoAdapterCardBinding
 import com.bumptech.glide.Glide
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
```diff
@@ -1,81 +1,101 @@
...
-            return if (oldItem is EventPhoto.Local && newItem is EventPhoto.Local) {
-                oldItem.uri == newItem.uri
-            } else if (oldItem is EventPhoto.Remote && newItem is EventPhoto.Remote) {
-                oldItem.photoUrl == newItem.photoUrl
+        override fun areItemsTheSame(oldItem: UiEventPhoto, newItem: UiEventPhoto): Boolean {
+            return if (oldItem is UiEventPhoto.Photo && newItem is UiEventPhoto.Photo) {
+                oldItem.eventPhoto.key == newItem.eventPhoto.key
+            } else if (oldItem is UiEventPhoto.AddPhoto && newItem is UiEventPhoto.AddPhoto) {
+                oldItem == newItem
             } else false
```
philipplackner, line 29: Shouldn't it be enough to replace this with `else oldItem == newItem`?

app/src/main/java/com/andrew/tasky/agenda/presentation/adapters/PhotoItemAdapter.kt
```diff
@@ -1,81 +1,101 @@
...
+                ) {
+                    oldItem.eventPhoto.uri == newItem.eventPhoto.uri
+                } else if (oldItem.eventPhoto is EventPhoto.Remote &&
+                    newItem.eventPhoto is EventPhoto.Remote
+                ) {
+                    oldItem.eventPhoto.photoUrl == newItem.eventPhoto.photoUrl
+                } else false
+            } else oldItem is UiEventPhoto.AddPhoto && newItem is UiEventPhoto.AddPhoto
+        }
+    }
```
philipplackner, line 45: Is every subclass of `UiEventPhoto` a data class? If so, you could just `return oldItem == newItem`

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailFragment.kt
```diff
@@ -246,37 +188,38 @@ class EventDetailFragment : Fragment(R.layout.fragment_event_detail) {
...
-
-                    attendeesLayout.addAttendeeButton.isVisible = isEditing
-                }
-
                 reminderLayout.reminderTextView.isEnabled = isEditing
                 reminderLayout.reminderButton.isEnabled = isEditing
                 reminderLayout.reminderButton.isVisible = isEditing
             }
 
+            collectLatestLifecycleFlow(viewModel.isNotAttendeeAndIsEditing) {
```
philipplackner, line 196: I'd usually rephrase such booleans to non-negated ones. In this case, how about `isCreatorEditing`? Creates less cognitive load

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailFragment.kt
```diff
@@ -297,17 +240,100 @@ class EventDetailFragment : Fragment(R.layout.fragment_event_detail) {
...
                 photoAdapter.submitList(photoList)
 
-                addPhotoLayout.addPhotoLayout.isVisible = !isAttendee || photoList.isNotEmpty()
                 addPhotoLayout.addPhotoPlusSign.isVisible = photoList.isEmpty()
                 addPhotoLayout.addPhotoTextView.isVisible = photoList.isEmpty()
                 addPhotoLayout.photosTextView.isVisible = photoList.isNotEmpty()
                 addPhotoLayout.photosRecyclerView.isVisible = photoList.isNotEmpty()
             }
 
+            collectLatestLifecycleFlow(viewModel.isNotAttendeeAndPhotosIsEmpty) {
```
philipplackner, line 332: Same here

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -95,6 +102,17 @@ class EventDetailViewModel @Inject constructor(
         _photo.value = photoList
     }
 
+    val uiEventPhotos = photos.map { photos ->
+        photos.map { photo ->
+            UiEventPhoto.Photo(photo)
+        }.toMutableList<UiEventPhoto>()
+            .apply {
+                if ((size in 1..9) && (!isAttendee.value)) {
+                    add(size, UiEventPhoto.AddPhoto)
```
philipplackner, line 111: If you just append at the end, you can omit the index where you want to insert

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -95,6 +102,17 @@ class EventDetailViewModel @Inject constructor(
         _photo.value = photoList
     }
 
+    val uiEventPhotos = photos.map { photos ->
+        photos.map { photo ->
+            UiEventPhoto.Photo(photo)
+        }.toMutableList<UiEventPhoto>()
+            .apply {
+                if ((size in 1..9) && (!isAttendee.value)) {
```
philipplackner, line 110: Avoid calling `.value` in a flow operator like here. That's very vulnerable to race conditions. In this case it's unlikely, but imagine `isAttendee` is a frequently changing flow. You could jump into this flow block here, read the first value of `isAttendee.value` and if it then changes right after the flow block will still execute with the previously read value. Instead you would need to use combine:
```
val uiEventPhotos = photos.combine(isAttendee) { photos, isAttendee ->
    // here you can now access isAttendee safely since it will be retriggered on every change
}
```

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -123,10 +141,12 @@ class EventDetailViewModel @Inject constructor(
...
-    val isAttending = _isAttending.asStateFlow()
-    fun switchAttendingStatus() {
-        _isAttending.value = !isAttending.value
+    val isNotAttendeeAndIsEditing = combine(isAttendee, isInEditMode) { isAttendee, isEditing ->
+        !isAttendee && isEditing
+    }
+    val isNotAttendeeAndPhotosIsEmpty = combine(isAttendee, uiEventPhotos) {
+        isAttendee, photos ->
+        !isAttendee && photos.isEmpty()
     }
```
philipplackner, line 150: I think you'd want to use StateFlows here right? Since you're observing these values in the UI, so you can cache the results

philipplackner (approved):

PR #43, opened 2023-01-14
Refactors Agenda Item into sealed interface

andrewhamilton0:
-Refactors Agenda Item into sealed interface

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
@@ -3,7 +3,6 @@ package com.andrew.tasky.agenda.data.agenda
 import com.andrew.tasky.agenda.data.util.ReminderTimeConversion
 import com.andrew.tasky.agenda.domain.AgendaRepository
 import com.andrew.tasky.agenda.domain.models.AgendaItem
-import com.andrew.tasky.agenda.util.AgendaItemType
 import com.andrew.tasky.auth.data.AuthResult
 import java.time.LocalDateTime
 import retrofit2.HttpException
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -18,9 +17,8 @@ class AgendaRepositoryImpl(
         return try {
             val response = api.getAgendaItems(timezone = timezone, time = time)
             val tasks = response.taskDtos.map { task ->
-                AgendaItem(
-                    apiId = task.id,
-                    type = AgendaItemType.TASK,
+                AgendaItem.Task(
+                    id = task.id.toInt(),
```
philipplackner, line 21: Why are these int IDs? You don't need separate IDs for local DB and API, you can just use the unique API IDs as primary key

app/src/main/java/com/andrew/tasky/agenda/domain/models/AgendaItem.kt
```diff
@@ -2,27 +2,44 @@ package com.andrew.tasky.agenda.domain.models
...
-    val startDateAndTime: LocalDateTime,
-    val endDateAndTime: LocalDateTime? = null,
-    val reminderTime: ReminderTime,
-    val photos: List<EventPhoto>? = emptyList(),
-    val isAttendee: Boolean? = null,
-    val attendees: List<Attendee>? = emptyList(),
-    val host: String? = null
-) : Serializable
+    @Entity
+    data class Event(
```
philipplackner, line 12: Here you're coupling domain with data related dependencies (in this case Room and its annotations). The domain layer should be isolated from all kinds of APIs and databases and it should never know where data comes from. That decision lies in the data layer.
So in your case, these entities would need to be moved to your data layer. Then you will have a mapper function which maps these entities to a corresponding domain model which doesn't have any Room or JSON annotations. The UI then uses these domain models to display stuff.

philipplackner (approved):

PR #44, opened 2023-01-16
Adds Mapper Functions

andrewhamilton0:
-Adds Mapper Functions
-Adds Reminder and Event Databases and Daos
-Auth Result Function

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/event/EventDao.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/event/EventDao.kt
@@ -1,0 +1,18 @@
+package com.andrew.tasky.agenda.data.event
+
+import androidx.room.*
+import com.andrew.tasky.agenda.domain.models.AgendaItem
+import kotlinx.coroutines.flow.Flow
+
+@Dao
+interface EventDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun upsert(event: AgendaItem.Event): Long
+
+    @Query("SELECT * FROM eventdto")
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/event/EventDao.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.andrew.tasky.agenda.domain.models.AgendaItem
+import kotlinx.coroutines.flow.Flow
+
+@Dao
+interface EventDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun upsert(event: AgendaItem.Event): Long
+
+    @Query("SELECT * FROM eventdto")
```
philipplackner, line 13: A DTO is usually a network related object (that represents JSON payload). I'd call database models simply entities. So Instead of EventDto, this would be EventEntity here

app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderDao.kt
```diff
@@ -0,0 +1,19 @@
...
+
+interface ReminderDao {
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun upsert(event: AgendaItem.Reminder): Long
+
+    @Query("SELECT * FROM ")
+    fun getEvents(): Flow<List<AgendaItem.Event>>
+
+    @Delete
+    suspend fun deleteEvent(event: AgendaItem.Reminder)
```
philipplackner, line 18: These should be reminders right?

philipplackner (approved):

PR #45, opened 2023-01-20
Adds Upload create reminder workmanager

andrewhamilton0:


app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderRepositoryImpl.kt
```diff
@@ -1,18 +1,41 @@
...
+                    .build()
+            ).setInputData(
+                workDataOf(
+                    "CREATE_REMINDER" to reminderDtoAsJson
+                )
+            )
+            .build()
+        WorkManager
+            .getInstance(appContext)
+            .enqueue(uploadReminderRequest)
```
philipplackner, line 35: For creating reminders WorkManager shouldn't be necessary since it's not a longer thing, it's just one network call that takes maybe a second. WorkManager makes sense for example when uploading files (like photos for an event) where there's a higher chance that the user goes away or something goes wrong

app/src/main/java/com/andrew/tasky/agenda/data/reminder/UploadReminderWorker.kt
```diff
@@ -0,0 +1,47 @@
...
+@HiltWorker
+class UploadReminderWorker @AssistedInject constructor(
+    @Assisted private val appContext: Context,
+    @Assisted private val workerParams: WorkerParameters,
+    private val api: ReminderApi,
+) : CoroutineWorker(appContext, workerParams) {
+
+    override suspend fun getForegroundInfo(): ForegroundInfo {
+        return super.getForegroundInfo()
+    }
```
philipplackner, line 24: This is not necessary I assume?

app/src/main/java/com/andrew/tasky/agenda/data/reminder/UploadReminderWorker.kt
```diff
@@ -0,0 +1,47 @@
...
+                Result.success()
+            }
+            is AuthResult.Unauthorized -> {
+                Log.e("REMINDER AUTH Result", "Unauthorized")
+                Result.failure()
+            }
+            is AuthResult.UnknownError -> {
+                Log.e("REMINDER AUTH Result", "Unknown Error")
+                Result.failure()
+            }
```
philipplackner, line 44: One cool thing about WorkManager is that you can also return `Result.retry()` for example when the server returned an error or so (maybe a temporary server side error).

app/src/main/java/com/andrew/tasky/agenda/domain/models/AgendaItem.kt
```diff
@@ -18,7 +19,7 @@ sealed interface AgendaItem : Serializable {
         val isAttendee: Boolean,
         val attendees: List<Attendee>? = emptyList(),
         val host: String
-    ) : AgendaItem
+    ) : AgendaItem, Serializable
```
philipplackner, line 22: I'd prefer `Parcelable` over `Serializable` which is just faster. You can use the Gradle Parcelize plugin to easily create parcelables simply by adding the `@Parcelize` annotation

philipplackner (approved):

PR #46, opened 2023-01-23
Adds some api functions for reminder

andrewhamilton0:
-Adds some api functions for reminder
-Adds get agenda item api function


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -1,35 +1,60 @@
...
+                reminderEntity.toReminder()
             }
-            response.reminderEntities.map { reminder ->
-                reminderDb.getReminderDao().upsert(reminder)
-            }
-            // TODO Make event and task db connect to api
         }
+        return localReminders
+        // Todo return task and event agenda items
+    }
```
philipplackner, line 27: If all you do is create a variable and then return it, I'd just return it right away, since there's no gain in having it. Variables make sense if you need them multiple times or they make something more readable, but here the function name kinda tells you what it represents :)

app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -1,35 +1,60 @@
...
+
+    override suspend fun syncAgendaItems(): AuthResult<Unit> {
+        val reminderDeleteIds = db.getReminderDao().getModifiedReminders().filter {
+            it.modifiedType == ModifiedType.DELETE
+        }.map {
+            it.id
+        }
```
philipplackner, line 34: I love how you get better at using these higher order functions, great work 🙌

app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderDao.kt
```diff
@@ -6,11 +6,20 @@ import kotlinx.coroutines.flow.Flow
...
     fun getReminders(): Flow<List<ReminderEntity>>
 
     @Delete
     suspend fun deleteReminder(reminder: ReminderEntity)
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun upsertModifiedReminder(modifiedReminder: ModifiedReminderEntity): Long
+
+    @Query("SELECT * FROM ModifiedReminderEntity")
+    fun getModifiedReminders(): List<ModifiedReminderEntity>
```
philipplackner, line 21: Should be `suspend` as well right? 😄

PR #47, opened 2023-01-23
Edits some event data classes,

andrewhamilton0:
-Edits some event data classes,
-Removes files not being used


app/src/main/java/com/andrew/tasky/agenda/data/util/MapperFunctions.kt
```diff
@@ -5,16 +5,17 @@ import com.andrew.tasky.agenda.data.reminder.ReminderDto
...
 import java.time.ZoneOffset
 
 fun TaskDto.toTask(): AgendaItem.Task {
     return AgendaItem.Task(
         id = id,
-        isDone = isDone,
         title = title,
         description = description ?: "",
-        startDateAndTime = LocalDateTime.ofEpochSecond(time, 0, null),
+        startDateAndTime = LocalDateTime.ofEpochSecond(time, 0, ZoneOffset.UTC),
```
philipplackner, line 18: If I'm not wrong, you want to use the local user's time zone here instead of `UTC` right? `LocalDateTime` does not contain any time zone information and should represent the local time as the name says. Alternatively, you can use `ZonedDateTime` which does contain this info :)

app/src/main/java/com/andrew/tasky/agenda/data/util/MapperFunctions.kt
```diff
@@ -62,21 +63,45 @@ fun AgendaItem.Reminder.toReminderEntity(): ReminderEntity {
     )
 }
 
+fun ReminderEntity.toReminder(): AgendaItem.Reminder {
+    return AgendaItem.Reminder(
+        id = id,
+        isDone = isDone,
+        title = title,
+        description = description,
+        startDateAndTime = LocalDateTime.ofEpochSecond(time, 0, ZoneOffset.UTC),
```
philipplackner, line 72: Same comment as before

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/task_detail/TaskDetailViewModel.kt
```diff
@@ -84,12 +82,12 @@ class TaskDetailViewModel @Inject constructor(
...
-                    // remoteRepository.createTask(agendaItem) TODO
-                } else {
-                    // remoteRepository.updateTask(agendaItem) TODO
-                }
+                // localRepository.upsert(agendaItem)
+                // if (savedStateHandle.get <AgendaItem.Task("agendaItem")?.id == null) {
+                //    // remoteRepository.createTask(agendaItem) TODO
+                // } else {
+                //    // remoteRepository.updateTask(agendaItem) TODO
+                // }
```
philipplackner, line 90: Throw it away if you don't need it :) If you do need it again, you can use Git's commit history and compare the changes with this commit.

philipplackner (approved):

PR #48, opened 2023-01-26
Added sync reminder function

andrewhamilton0:
-Added sync reminder function
-Fixed time conversion for ZonedDateTime

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
@@ -2,22 +2,27 @@ package com.andrew.tasky.agenda.data.agenda
 
 import android.util.Log
 import com.andrew.tasky.agenda.data.database.AgendaDatabase
+import com.andrew.tasky.agenda.data.reminder.ReminderApi
 import com.andrew.tasky.agenda.data.util.*
 import com.andrew.tasky.agenda.domain.AgendaRepository
+import com.andrew.tasky.agenda.domain.ReminderRepository
 import com.andrew.tasky.agenda.domain.models.AgendaItem
 import com.andrew.tasky.auth.data.AuthResult
 import com.andrew.tasky.auth.util.getAuthResult
+import java.time.LocalDate
+import java.time.LocalTime
 import java.util.*
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 
 class AgendaRepositoryImpl(
-    private val api: AgendaApi,
+    private val agendaApi: AgendaApi,
+    private val reminderRepository: ReminderRepository,
     private val db: AgendaDatabase,
 ) : AgendaRepository {
 
     override fun getAgendaItems(): Flow<List<AgendaItem>> {
-        val localReminders = db.getReminderDao().getReminders().map {
+        val localReminders = db.getReminderDao().getRemindersAsFlow().map {
             it.map { reminderEntity ->
                 reminderEntity.toReminder()
             }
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -26,31 +31,60 @@ class AgendaRepositoryImpl(
...
                 }
-                remoteReminders?.map { reminder ->
-                    db.getReminderDao().upsertReminder(reminder.toReminderEntity())
+                results.data?.reminders?.map { reminderDto ->
+                    val localReminder = db.getReminderDao().getReminderById(reminderDto.id)
+                    db.getReminderDao().upsertReminder(
+                        reminderDto.toReminderEntity(
+                            // Android Studio says local reminder will never equal null,
+                            // but can if reminder doesn't exist in db
+                            isDone = localReminder.isDone ?: false
```
philipplackner, line 85: Android Studio is actually correct here I think. Is the return type of `getReminderById()` `Reminder?` or `Reminder`? If it's the latter, the function will throw an exception if the reminder doesn't exist.

app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderRepositoryImpl.kt
```diff
@@ -57,4 +59,48 @@ class ReminderRepositoryImpl @Inject constructor(
...
+                    "Unknown Error, could not create reminder"
+                )
+            }
+        }
+
+        val updateRemindersDtos = db.getReminderDao().getModifiedReminders().filter {
+            it.modifiedType == ModifiedType.UPDATE
+        }.map {
+            db.getReminderDao().getReminderById(it.id).toReminderDto()
+        }
```
philipplackner, line 89: The top and bottom call are very similar and use the same data, however you read from the DB twice. You can also make use of `groupBy` which groups a list of items based on a specific key:
```
val modifiedReminders = db.getReminderDao().getModifiedReminders().groupBy { it.modifiedType }

// Get created reminders
val createdReminders = modifiedReminders[ModifiedType.CREATE]

// Get updated reminders
val updatedReminders = modifiedReminders[ModifiedType.UPDATE]
```

app/src/main/java/com/andrew/tasky/agenda/data/util/ReminderTimeConversion.kt
```diff
@@ -39,21 +42,33 @@ object ReminderTimeConversion {
...
+                ZonedDateTime.of(
+                    startLocalDateTime.minusHours(6),
+                    TimeZone.getDefault().toZoneId()
+                ).toEpochSecond()
             ReminderTime.ONE_DAY_BEFORE ->
-                startTime.minusDays(1).toEpochSecond(ZoneOffset.UTC)
+                ZonedDateTime.of(
+                    startLocalDateTime.minusDays(1),
+                    TimeZone.getDefault().toZoneId()
+                ).toEpochSecond()
```
philipplackner, line 71: All these calls are very similar, could it make sense to create a helper function that just takes in the amount of time you want to subtract?
Also, you're returning the timestamp in seconds again instead of milliseconds

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/agenda/AgendaViewModel.kt
```diff
@@ -134,4 +136,17 @@ class AgendaViewModel@Inject constructor(
...
+            .setConstraints(
+                Constraints.Builder()
+                    .setRequiredNetworkType(NetworkType.CONNECTED)
+                    .build()
+            ).build()
+
+    init {
+        viewModelScope.launch { updateAgendaOfDate(date = dateSelected.value) }
+        workManager.enqueue(syncModifiedAgendaItemsWorkRequest)
+    }
```
philipplackner, line 151: Using WorkManager directly in a ViewModel prevents that it can be locally unit tested since WorkManager is an Android dependency. You could do this to solve this:
```
// In domain
interface AgendaItemScheduler {
    fun scheduleSync(minuteInterval: Int)
}

// In data
class AgendaItemSchedulerImpl(val workManager: WorkManager): AgendaItemScheduler {
    // Implement function
}

// In ViewModel
class AgendaViewModel(val scheduler: AgendaItemScheduler): ViewModel() {
    init {
        scheduler.scheduleSync(minuteInterval = 15)
    }
}
```

This way, you could pass another implementation of `AgendaItemScheduler` to your ViewModel in a test case which does not depend on WorkManager

philipplackner (approved):

PR #49, opened 2023-01-31
01/30/2023 combines two functions in agenda repo

andrewhamilton0:
-Adds zonedEpochMilli conversions
-Combines getAgendaItems function and updateAgendaOfDate

### DIFF
--- a/.idea/deploymentTargetDropDown.xml
+++ b/.idea/deploymentTargetDropDown.xml
@@ -1,18 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="deploymentTargetDropDown">
-    <targetSelectedWithDropDown>
-      <Target>
-        <type value="QUICK_BOOT_TARGET" />
-        <deviceKey>
-          <Key>
-            <type value="VIRTUAL_DEVICE_PATH" />
-            <value value="C:\Users\Andrew\.android\avd\Pixel_C_API_33.avd" />
-          </Key>
-        </deviceKey>
-      </Target>
-    </targetSelectedWithDropDown>
-    <timeTargetWasSelectedWithDropDown value="2023-01-31T00:05:11.485064500Z" />
     <runningDeviceTargetsSelectedWithDialog>
       <Target>
         <type value="RUNNING_DEVICE_TARGET" />


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -2,32 +2,93 @@ package com.andrew.tasky.agenda.data.agenda
...
 import java.util.*
+import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 
 class AgendaRepositoryImpl(
     private val agendaApi: AgendaApi,
     private val reminderRepository: ReminderRepository,
     private val db: AgendaDatabase,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
```
philipplackner, line 23: If it's a variable dispatcher, I wouldn't call it `ioDispatcher`, since that's not true anymore as soon as it's a different one

app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -2,32 +2,93 @@ package com.andrew.tasky.agenda.data.agenda
...
+            localDate.dayOfMonth
+        )
+        calendar.set(Calendar.HOUR_OF_DAY, 0)
+        calendar.set(Calendar.MINUTE, 0)
+        calendar.set(Calendar.SECOND, 0)
+        calendar.set(Calendar.MILLISECOND, 0)
+        val startEpochMilli = calendar.timeInMillis
+
+        calendar.add(Calendar.DATE, 1)
+        val endEpochMilli = calendar.timeInMillis
```
philipplackner, line 41: You don't really need `Calendar` for anything, it's just a lot of boilerplate code here. You can do this all with `LocalDateTime`/ `ZonedDateTime` more convenient and with less code

app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -2,32 +2,93 @@ package com.andrew.tasky.agenda.data.agenda
...
+
+        CoroutineScope(ioDispatcher).launch {
```
philipplackner, line 43: I'd avoid launching coroutines outside of ViewModels. These are uncontrollable by the ViewModel. Imagine, your VM calls this function and the user navigates away right after. Then this call shouldn't resume, which it would with a custom scope here though.
If the user navigates back `viewModelScope` is cleared and therefore all suspending calls in it would be cancelled. Long story short, only expose suspend functions unless in a ViewModel or you're creating an object that has an actual lifecycle different than the ViewModel (which is not the case for repos)

app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -2,32 +2,93 @@ package com.andrew.tasky.agenda.data.agenda
...
+                        startEpochMilli = startEpochMilli,
+                        endEpochMilli = endEpochMilli
+                    )
+                    localReminders.map { localReminder ->
+                        val remoteReminderIds =
+                            results.data?.reminders?.map { it.id } ?: emptyList()
+                        if (!remoteReminderIds.contains(localReminder.id)) {
+                            db.getReminderDao().deleteReminder(localReminder)
+                        }
+                    }
```
philipplackner, line 66: Does this outer map do anything here? You're not assigning the result of mapping to anything. Looks like all you want here is `forEach`.
Also the content can be shortened to this:
```
val containsLocalId = results.data?.reminders?.any { it.id == localReminder.id } == true
if(!containsLocalId) {
    // Do your stuff
}
```

philipplackner (approved):

PR #50, opened 2023-02-02
Adds Task Entity

andrewhamilton0:
-Adds Task Entity
-Adds Task Mapper Functions
-Makes separate file for mapper functions of each type, moved reminder mapper function to respective file
-Adds Modified Task Entity
-Used ZoneLocalTime at beginning of day rather than calendar instance
-Got rid of coroutine in get agendaItem function in AgendaRepositoryImpl.kt and made it a suspend function


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -20,66 +22,52 @@ class AgendaRepositoryImpl(
     private val agendaApi: AgendaApi,
     private val reminderRepository: ReminderRepository,
     private val db: AgendaDatabase,
-    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
 ) : AgendaRepository {
 
-    override fun getAgendaItems(localDate: LocalDate): Flow<List<AgendaItem>> {
+    override suspend fun getAgendaItems(localDate: LocalDate): Flow<List<AgendaItem>> {
```
philipplackner, line 27: To answer your question in WhatsApp, the issue lies in this function. When implementing a local cache you mainly have 2 options:
1. In this function you directly return a flow with the `flow {  }` builder and the first thing you do is to emit the local agenda items from the DB. Then you load them remotely and insert them in the cache. Finally, you emit the local items from the cache again. That way, the UI will first receive the items from the DB and then after the items have been fetched from the API.
2. You write two separate functions. One function does nothing else than returning the flow from the DB which will emit as soon something in the table changes. Then the other function is only responsible for updating the DB with the remote items. Since your UI will always get the latest changes from the first function, it will also directly get them when another function inserts something in the DB.

Approach 1 has the advantage that you have the caching logic at one place and approach 2 has the advantage that you can get the cache without making a network call.

app/src/main/java/com/andrew/tasky/agenda/data/task/TaskDao.kt
```diff
@@ -1,3 +1,42 @@
...
+    fun getTasksOfDateFlow(
+        startEpochMilli: Long,
+        endEpochMilli: Long
+    ): Flow<List<TaskEntity>>
+
+    @Query("SELECT * FROM TaskEntity WHERE time BETWEEN :startEpochMilli AND :endEpochMilli")
+    suspend fun getTasksOfDate(
+        startEpochMilli: Long,
+        endEpochMilli: Long
+    ): List<TaskEntity>
```
philipplackner, line 23: It's enough to just have the flow function, the second one is equivalent to `flow.first()`

PR #52, opened 2023-02-06
Sets up create event in event repository

andrewhamilton0:
-Sets up create event in event repository
-Changes get agenda item to emit agenda items before network call

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
@@ -17,7 +17,6 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.LocalTime
 import java.util.*
-import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.*
 
 class AgendaRepositoryImpl(
@@ -27,10 +26,36 @@ class AgendaRepositoryImpl(
     private val db: AgendaDatabase,
 ) : AgendaRepository {
 
-    override suspend fun getAgendaItems(localDate: LocalDate): Flow<List<AgendaItem>> {
-
+    override suspend fun getAgendaItems(localDate: LocalDate) = flow<Flow<List<AgendaItem>>> {
         val startEpochMilli = localDateTimeToZonedEpochMilli(localDate.atStartOfDay())
         val endEpochMilli = localDateTimeToZonedEpochMilli(localDate.atStartOfDay().plusDays(1))
+
+        val reminders = db.getReminderDao().getRemindersOfDate(
+            startEpochMilli = startEpochMilli,
+            endEpochMilli = endEpochMilli
+        ).map {
+            it.map { reminderEntity ->
+                reminderEntity.toReminder()
+            }
+        }
+        val tasks = db.getTaskDao().getTasksOfDate(
+            startEpochMilli = startEpochMilli,
+            endEpochMilli = endEpochMilli
+        ).map {
+            it.map { taskEntity ->
+                taskEntity.toTask()
+            }
+        }
+
+        emit(
+            reminders.combine(tasks) { reminderFlow, taskFlow ->
+                reminderFlow + taskFlow
+            }.map {
+                it.sortedBy { agendaItem ->
+                    agendaItem.startDateAndTime
+                }
+            }
+        )
 
         syncAgendaItems()
         when (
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -27,11 +26,37 @@ class AgendaRepositoryImpl(
...
+
+        emit(
+            reminders.combine(tasks) { reminderFlow, taskFlow ->
+                reminderFlow + taskFlow
+            }.map {
+                it.sortedBy { agendaItem ->
+                    agendaItem.startDateAndTime
+                }
+            }
+        )
```
philipplackner, line 58: In your flow you're emitting flows, instead of just `List<AgendaItem>`. It's comparable to a list of lists and not what you want here.
Instead you should return a `Flow<List<AgendaItem>>` here and in the last line you can use `emitAll(reminders.combine(...) { ... })` which will listen to that flow and send all emissions in the flow you return

app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -132,22 +129,28 @@ class AgendaRepositoryImpl(
...
         }
-        val syncAgendaRequest = SyncAgendaRequest(emptyList(), taskDeleteIds, reminderDeleteIds)
-        when (getAuthResult { agendaApi.syncAgendaItems(syncAgendaRequest) }) {
-            is AuthResult.Authorized -> {
-                reminderDeleteIds.map {
-                    db.getReminderDao().deleteModifiedReminderById(it)
+
+        if (reminderDeleteIds.isNotEmpty() || taskDeleteIds.isNotEmpty()) {
+            val syncAgendaRequest = SyncAgendaRequest(emptyList(), taskDeleteIds, reminderDeleteIds)
+            when (getAuthResult { agendaApi.syncAgendaItems(syncAgendaRequest) }) {
```
philipplackner, line 135: I'd personally avoid putting such longer functions in a single if or when statement, as it's harder to read then. I'd instead save the result of `getAuthResult` in a variable here and then use that in when. That way you can also apply different formatting

app/src/main/java/com/andrew/tasky/agenda/data/event/EventMapperFunctions.kt
```diff
@@ -0,0 +1,22 @@
...
+        id = id,
+        title = title,
+        description = description,
+        from = localDateTimeToZonedEpochMilli(startDateAndTime),
+        to = localDateTimeToZonedEpochMilli(endDateAndTime),
+        remindAt = ReminderTimeConversion.toEpochMilli(
+            reminderTime = reminderTime,
+            startLocalDateTime = startDateAndTime
+        ),
+        attendeeIds = attendees?.map {
```
philipplackner, line 18: Why can `attendees` be null here? If there are no attendees, I'd simply use an empty list because that says exactly that. I'd use null values for all other types where it's not clear that a default value indicates a non-existent value like for Strings

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -1,18 +1,43 @@
...
+                        is EventPhoto.Local -> {
+                            MultipartBody.Part
+                                .create(eventPhoto.uri.toFile().readBytes().toRequestBody())
+                        }
+                        is EventPhoto.Remote -> {
+                            MultipartBody.Part
+                                .create(eventPhoto.photoUrl.encodeToByteArray().toRequestBody())
+                        }
+                    }
+                } ?: emptyList()
```
philipplackner, line 38: 1. You're correctly attaching the photos and the JSON, however you don't specify any names for it. The API needs to know what each attachment represents, for which you can attach the key like `create_event_request`. For the photos it's `photo0`, `photo1`, ..., `photo9`
2. There's no point in uploading remote photos again as they are already on the server (or here for creating events there's no scenario where there would even be a remote photo). You can therefore safely just filter for local photos using `event.photos.filterIsInstance<EventPhoto.Local>`

app/src/main/java/com/andrew/tasky/agenda/domain/AgendaRepository.kt
```diff
@@ -5,6 +5,6 @@ import java.time.LocalDate
 import kotlinx.coroutines.flow.Flow
 
 interface AgendaRepository {
-    suspend fun getAgendaItems(localDate: LocalDate): Flow<List<AgendaItem>>
+    suspend fun getAgendaItems(localDate: LocalDate): Flow<Flow<List<AgendaItem>>>
```
philipplackner, line 8: Same thing as in the first comment here

philipplackner (approved):

PR #53, opened 2023-02-07
02/07/2023 creates logout function

andrewhamilton0:
-Sets up logout function, deletes agenda tables and shared prefs when logging out
-Sets up shared prefs username and id


andrewhamilton0:
Closed because forgot to merge previous PR before merging

PR #54, opened 2023-02-07
02/07/2023 creates logout function 1

andrewhamilton0:
Adds logout function
Adds username and user id to prefs


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -153,4 +153,11 @@ class AgendaRepositoryImpl(
             // Todo sync events
         }
     }
+
+    override suspend fun deleteAllAgendaTables() {
+        db.getReminderDao().deleteModifiedReminderDb()
+        db.getReminderDao().deleteReminderDb()
+        db.getTaskDao().deleteModifiedTaskDb()
+        db.getTaskDao().deleteTaskDb()
+    }
```
philipplackner, line 162: You can also just do `db.clearAllTables()`, otherwise you might forget to extend this function when you implement more agenda items in future. Of course this only works if you know you'll always want to clear the whole DB, so if there was some other unrelated DB schema, that could go in its own DB.

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/agenda/AgendaViewModel.kt
```diff
@@ -130,6 +143,17 @@ class AgendaViewModel@Inject constructor(
         }
     }
 
+    private val logoutResultChannel = Channel<AuthResult<Unit>>()
+    val logoutResults = logoutResultChannel.receiveAsFlow()
+
+    fun logout() {
+        viewModelScope.launch {
+            withContext(NonCancellable) {
+                logoutResultChannel.send(authRepository.logout())
```
philipplackner, line 152: Is the result here really necessary? What would you do if logging out fails, in case the user is offline for example? They'd still want to logout, so do you really care about an error?

philipplackner (approved):

PR #55, opened 2023-02-09
Adds event update request

andrewhamilton0:
-Changes getAgendaItems from AgendaRepositoryImpl.kt to use emitAll
-Adds event update request
-Changes api.create to use .createFormData in EventRepositoryImpl.kt

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
@@ -26,7 +26,7 @@ class AgendaRepositoryImpl(
     private val db: AgendaDatabase,
 ) : AgendaRepository {
 
-    override suspend fun getAgendaItems(localDate: LocalDate) = flow<Flow<List<AgendaItem>>> {
+    override suspend fun getAgendaItems(localDate: LocalDate) = flow<List<AgendaItem>> {
         val startEpochMilli = localDateTimeToZonedEpochMilli(localDate.atStartOfDay())
         val endEpochMilli = localDateTimeToZonedEpochMilli(localDate.atStartOfDay().plusDays(1))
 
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -26,7 +26,7 @@ class AgendaRepositoryImpl(
     private val db: AgendaDatabase,
 ) : AgendaRepository {
 
-    override suspend fun getAgendaItems(localDate: LocalDate) = flow<Flow<List<AgendaItem>>> {
+    override suspend fun getAgendaItems(localDate: LocalDate) = flow<List<AgendaItem>> {
```
philipplackner, line 29: There's no point in returning a flow and directly calling .first(). It will only return the first emission. The point of a flow is to be able to listen to changes at different times, so you also want to return the flow and not the result of it :)

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -20,24 +21,35 @@ class EventRepositoryImpl @Inject constructor(
...
+                        name = "create_event_request",
+                        value = Json.encodeToString(event.toCreateEventRequest())
+                    ),
+                photoData = event.photos.filterIsInstance<EventPhoto.Local>().mapIndexed {
+                    index, eventPhoto ->
+                    MultipartBody.Part
+                        .createFormData(
+                            name = "photo$index",
+                            filename = eventPhoto.key,
+                            body = eventPhoto.uri.toFile().readBytes().toRequestBody()
```
philipplackner, line 37: Have you tried this? I think you can only convert file uris that start with `file://` to a file. If you pick an image from the gallery it's as content uri from which you get the byte array using `contentResolver`

PR #57, opened 2023-02-11
Changed isAttendee to isCreator in Event

andrewhamilton0:
-Changed isAttendee to isCreator in Event
-Added some converters back
-Changed dateTimeConversions to extension functions -Added some event mapper functions
-Added ByteArray HelperFunctions


andrewhamilton0:
Removed Converters Functions and Event Mappers to shorten PR

app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -46,10 +47,21 @@ class AgendaRepositoryImpl(
...
+                eventEntity ->
+                eventEntity.toEvent()
+            }
+        }
 
         emit(
             reminders.combine(tasks) { reminderFlow, taskFlow ->
                 reminderFlow + taskFlow
+            }.combine(events) { reminderTaskFlow, eventsFlow ->
+                reminderTaskFlow + eventsFlow
```
philipplackner, line 64: I assume `reminderFlow` etc. are lists of agenda items right? Then why call them flow and not `reminders`?

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -30,25 +35,38 @@ class EventRepositoryImpl @Inject constructor(
                     ),
                 photoData = event.photos.filterIsInstance<EventPhoto.Local>().mapIndexed {
                     index, eventPhoto ->
+                    val imageByte = withContext(Dispatchers.IO) {
+                        appContext.contentResolver.openInputStream(eventPhoto.uri).use {
+                            it!!.readBytes()
+                        }
+                    }
+                    val compressedImageByte = compressByteArrayToSize(imageByte, 1000000)
```
philipplackner, line 43: I'd create a separate class for reading bytes from a content uri to follow the single source of truth principle. That allows you to easily reuse this logic and it's easier to find and change. I created a `UriByteConverter` class or so for that

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -44,10 +44,10 @@ class EventDetailViewModel @Inject constructor(
...
 
-    private val _isAttendee = MutableStateFlow(false)
-    val isAttendee = _isAttendee.asStateFlow()
-    fun setIsAttendee(isAttendee: Boolean) {
-        _isAttendee.value = isAttendee
+    private val _isCreator = MutableStateFlow(true)
+    val isCreator = _isCreator.asStateFlow()
+    private fun setIsCreator(isCreator: Boolean) {
+        _isCreator.value = isCreator
     }
```
philipplackner, line 51: Why is this function needed? It's private, so only the ViewModel can access it, but why shouldn't it change `_isCreator` directly?

PR #58, opened 2023-02-11
Adds Converters Functions back

andrewhamilton0:
-Adds Converters Functions back
- Adds more Event Mapper Functions


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -57,16 +57,17 @@ class AgendaRepositoryImpl(
...
+                _reminders + _tasks
+            }.combine(events) { _remindersAndTasks, _events ->
+                _remindersAndTasks + _events
             }.map {
                 it.sortedBy { agendaItem ->
                     agendaItem.startDateAndTime
                 }
-            }.first()
+            }
         )
```
philipplackner, line 71: The reason that nothing is happening in your flow after this `emitAll()` is that this function will block the flow until all emissions are sent into it. However, since the DB flows never complete (because there could always be new emissions when the table changes), this `emitAll()` will also never complete.
I'd get rid of the extra complexity here and have a simple function that just updates the local DB after getting the agenda items from the network. Then you don't need any flow logic except for just returning the flow from the DB as it is. So something like this:
```
suspend fun updateAgendaItemCache(...) {
    // Retrofit call to get agenda items for the day
    ...
    // Insert all items in local DB
}

// This flow that is returned here will emit new items every time the updateAgendaItemCache function from above is called
fun getAgendaItems(...): Flow<List<AgendaItem>> {
    return combine(dao.getEvents(...), dao.getTasks(...), dao.getReminders(...)) { events, tasks, reminders ->
        (events + tasks + reminders).sortedBy { it.dateAndTime }
    }
}
```

app/src/main/java/com/andrew/tasky/agenda/data/util/UriByteConverter.kt
```diff
@@ -0,0 +1,27 @@
...
+        var quality = 100
+        var result = imageByte
+        while (result.size > targetSize && quality > 0) {
+            val stream = ByteArrayOutputStream()
+            bitmap.compress(Bitmap.CompressFormat.JPEG, quality, stream)
+            result = stream.toByteArray()
+            quality -= 5
+        }
+        return result
+    }
```
philipplackner, line 26: The code for compressing the image should also rather go in a separate class, for example `BitmapCompressor` or so. In your case that's quite a side effect of the `uriToByteArray` function. Imagine that function would come from a library and would just silently compress the read image without you being able to change that behavior. That's one reason why we stick to the single-responsibility principle, so that we're really flexible with how we combine different classes.

PR #59, opened 2023-02-13
Adds event photo dtos in order to store into Room Database

andrewhamilton0:
-Adds event photo dtos in order to store into Room Database
-Adds Converters.kt for photos
-Adds some mappers for eventPhotoDtos

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/domain/models/EventPhoto.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/domain/models/EventPhoto.kt
@@ -1,17 +1,20 @@
 package com.andrew.tasky.agenda.domain.models
 
-import android.net.Uri
+import android.os.Parcelable
 import java.util.UUID
+import kotlinx.parcelize.Parcelize
 
 sealed class EventPhoto(open val key: String) {
 
+    @Parcelize
     data class Remote(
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/andrew/tasky/agenda/domain/models/EventPhoto.kt
```diff
@@ -1,17 +1,20 @@
...
+import kotlinx.parcelize.Parcelize
 
 sealed class EventPhoto(open val key: String) {
 
+    @Parcelize
     data class Remote(
         override val key: String = UUID.randomUUID().toString(),
         val photoUrl: String,
-    ) : EventPhoto(key = key)
+    ) : EventPhoto(key = key), Parcelable
```
philipplackner, line 13: You need to be a bit careful with default arguments and parcelables. I'm not sure if this is an issue here and I would definitely try if it works first, but I remember that I had an issue with this case. AFAIK, if a Parcelable is passed somewhere, Android will not preserve the same instance you pass, instead it could be that it recreates a new instance where you extract it. This could cause that a new ID is generated if none was passed first.
But again, it's just in my memory that I had some issues with that, but it might not apply here..

philipplackner (approved):

PR #61, opened 2023-02-16
Combines create and update event function into one upsert function

andrewhamilton0:
-Combines create and update event function into one upsert function. This gives the data layer the decision to create or update an event and also gets rid of glitch where if user creates and updates an event in offline mode, it will only try to send an updated event instead of creating it remotely

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
@@ -26,112 +26,145 @@ class EventRepositoryImpl @Inject constructor(
     private val ioDispatcher: CoroutineDispatcher
 ) : EventRepository {
 
-    override suspend fun createEvent(event: AgendaItem.Event) {
-        db.getEventDao().upsertEvent(event.toEventEntity())
-        val result = getAuthResult {
-            api.createEvent(
-                eventData = MultipartBody.Part
-                    .createFormData(
-                        name = "create_event_request",
-                        value = Json.encodeToString(event.toCreateEventRequest())
-                    ),
-                photoData = event.photos.filterIsInstance<EventPhoto.Local>().mapIndexed {
-                    index, eventPhoto ->
-                    val uriByteConverter = UriByteConverter(
-                        appContext = appContext,
-                        ioDispatcher = ioDispatcher
-                    )
-                    val imageByte = uriByteConverter.uriToByteArray(
-                        uri = Uri.parse(eventPhoto.uri),
-                    )
-
-                    val compressedImageByte = BitmapCompressor().compressByteArray(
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -26,112 +26,145 @@ class EventRepositoryImpl @Inject constructor(
...
-                            filename = eventPhoto.key,
-                            body = compressedImageByte.toRequestBody()
+                            name = "create_event_request",
+                            value = Json.encodeToString(event.toCreateEventRequest())
+                        ),
+                    photoData = event.photos.filterIsInstance<EventPhoto.Local>().mapIndexed {
+                        index, eventPhoto ->
+                        val uriByteConverter = UriByteConverter(
+                            appContext = appContext,
+                            ioDispatcher = ioDispatcher
```
philipplackner, line 45: You don't really need to pass the dispatcher to classes like this. There is a use case for injecting dispatchers in ViewModels, since you might want to replace them for local unit tests (in which the default Main dispatcher is not available), but data related classes can't be locally unit tested anyways since they usually contain Android related dependencies. So feel free to just hardcode the dispatcher as `withContext(Dispatchers.IO)`

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -26,112 +26,145 @@ class EventRepositoryImpl @Inject constructor(
...
+                        )
+                        val imageByte = uriByteConverter.uriToByteArray(
+                            uri = Uri.parse(eventPhoto.uri),
+                        )
+
+                        val compressedImageByte = BitmapCompressor().compressByteArray(
```
philipplackner, line 51: If your class does not need constructor parameters, you can also make it an `object` 👍

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -26,112 +26,145 @@ class EventRepositoryImpl @Inject constructor(
...
+                }
+                is AuthResult.UnknownError -> {
+                    db.getEventDao().upsertModifiedEvent(
+                        ModifiedEventEntity(
+                            id = event.id,
+                            modifiedType = ModifiedType.UPDATE
                         )
                     )
+                    AuthResult.Unauthorized()
                 }
```
philipplackner, line 155: If multiple branches of a when expression contain the same code, you can combine them:
```
is AuthResult.Unauthorized,
is AuthResult.UnknownError -> {
    // ...
}
```

philipplackner (approved):

PR #62, opened 2023-02-18
Added add attendee functionality

andrewhamilton0:
-Added add attendee functionality
-Disables add attendee button if email is invalid
-Adds UiText sealed class
-If there is a token in prefs, then main activity directs fragment to agenda screen

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
@@ -3,12 +3,15 @@ package com.andrew.tasky.agenda.presentation.screens.event_detail
 import android.content.SharedPreferences
 import android.net.Uri
 import androidx.lifecycle.*
+import com.andrew.tasky.R
 import com.andrew.tasky.agenda.domain.EventRepository
 import com.andrew.tasky.agenda.domain.models.AgendaItem
 import com.andrew.tasky.agenda.domain.models.Attendee
 import com.andrew.tasky.agenda.domain.models.EventPhoto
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -109,11 +113,62 @@ class EventDetailViewModel @Inject constructor(
...
 
     private val _attendees = MutableStateFlow(listOf<Attendee>())
-    val attendees = _attendees.asStateFlow()
-    fun addAttendee(attendee: Attendee) {
-        _attendees.value += attendee
+    private val attendees = _attendees.asStateFlow()
+
+    fun addAttendee(email: String) {
+        viewModelScope.launch {
+            withContext(NonCancellable) {
```
philipplackner, line 120: Any reason for this? Are you popping the backstack right after?

app/src/main/java/com/andrew/tasky/core/MainActivity.kt
```diff
@@ -1,15 +1,31 @@
...
+import androidx.navigation.fragment.NavHostFragment
 import com.andrew.tasky.R
+import com.andrew.tasky.core.data.PrefsKeys
 import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
 
 @AndroidEntryPoint
 class MainActivity : AppCompatActivity() {
+    private lateinit var navController: NavController
+    @Inject lateinit var prefs: SharedPreferences
```
philipplackner, line 16: I'd avoid injecting stuff directly in the UI, I usually try to keep it as dumb as possible and only let the VM decide what the UI should do or display. So reading from prefs could happen in the VM here :)

PR #63, opened 2023-02-22
Added resource class

andrewhamilton0:
Added resource class, Modified GetAuth Results to GetResourceResult.kt, and Added Error Message Dto in order to get error messages from api


app/src/main/java/com/andrew/tasky/agenda/data/event/EventApi.kt
```diff
@@ -9,36 +11,36 @@ interface EventApi {
...
         @Part photoData: List<MultipartBody.Part>
-    ): EventDto
+    ): Response<EventDto>
 
     @GET("/event")
     suspend fun getEvent(
         @Query("eventId")
         eventId: String
-    ): EventDto
+    ): Response<EventDto>
```
philipplackner, line 20: Why this change?

app/src/main/java/com/andrew/tasky/auth/util/GetResourceResult.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlinx.coroutines.withContext
+import okhttp3.ResponseBody
+import okio.IOException
+import retrofit2.HttpException
+import retrofit2.Response
+
+suspend fun <T> getResourceResult(apiToBeCalled: suspend () -> Response<T>): Resource<T> {
+    return withContext(Dispatchers.IO) {
+        try {
+            val response = apiToBeCalled()
```
philipplackner, line 17: If you return a `Response` here, then afaik this doesn't throw an exception and instead includes the error in the response result, so the try catch isn't necessary if I remember correctly

PR #64, opened 2023-02-22
Adds resource class to auth

andrewhamilton0:


app/src/main/java/com/andrew/tasky/auth/data/LogoutWorker.kt
```diff
@@ -18,21 +19,17 @@ class LogoutWorker @AssistedInject constructor(
...
-                Log.e("Logout work request", "unauthorized")
-                Result.failure()
-            }
-            is AuthResult.UnknownError -> {
-                Log.e("Logout work request", "unknown error")
+            is Resource.Error -> {
+                Log.e("Logout work request", result.message ?: "unknown error")
+                // Should I have this set to failure or retry, will it retry forever if set to retry
+                // and never succeeds? Could this be an issue? I do have a policy that checks
+                // network connection before trying
```
philipplackner, line 29: Yes it will retry forever, but inside a worker there's a variable which is called something like `runAttemptCount` (or similar) which you can use to find out which attempt this is. For example:
```
if(runAttemptCount >= 3) {
    return Result.failure()
}
```

PR #65, opened 2023-02-23
Adds ui text to resource and GetResourceResult.kt classes

andrewhamilton0:
-Adds ui text to resource and GetResourceResult.kt classes

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderRepositoryImpl.kt
@@ -1,5 +1,6 @@
 package com.andrew.tasky.agenda.data.reminder
 
+import android.content.Context
 import android.util.Log
 import com.andrew.tasky.agenda.data.database.AgendaDatabase
 import com.andrew.tasky.agenda.data.util.ModifiedType
@@ -11,7 +12,8 @@ import javax.inject.Inject
 
 class ReminderRepositoryImpl @Inject constructor(
     private val db: AgendaDatabase,
-    private val api: ReminderApi
+    private val api: ReminderApi,
+    private val appContext: Context
 ) : ReminderRepository {
 
     override suspend fun createReminder(reminder: AgendaItem.Reminder) {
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderRepositoryImpl.kt
```diff
@@ -88,7 +90,7 @@ class ReminderRepositoryImpl @Inject constructor(
                     is Resource.Error -> {
                         Log.e(
                             "uploadCreateAndUpdateModifiedReminders error",
-                            result.message ?: "Unknown Error"
+                            result.message?.asString(appContext) ?: "Unknown Error"
```
philipplackner, line 93: For error logging, you could also extend `UiText` with a function which returns the value without context if it's a dynamic string (which error messages relevant for logs would usually be). What I mean:
```
// Inside UiText.kt
fun tryGetValue(): String? {
    return if(this is DynamicString) value else null
}
```
Then you don't need to pass the app context around in your repos just for the sake of logging. But this is optional, not really something you have to do

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -137,20 +133,13 @@ class EventDetailViewModel @Inject constructor(
...
-                                    _attendees.value += result.data.attendee
-                                } else {
-                                    attendeeToastMessageChannel.send(
-                                        UiText.Resource(
-                                            resId = R.string.attendee_already_added
-                                        )
-                                    )
-                                }
+                            val attendee = result.data.copy(isGoing = true)
+                            if (!attendees.value.contains(attendee)) {
```
philipplackner, line 137: It would probably make more sense here to check if there's already an attendee with the same ID rather than checking if there's an attendee that has exactly the same fields. With your implementation, if attendee A is marked as not going and it's already in the list, it would get added again here, since it would be marked as going so it's considered as a different attendee.
Instead you can do this:
```
val isAttendeePartOfEvent = attendees.value.any { it.id == result.data.id }
if(!isAttendeePartOfEvent) { ... }
```

PR #66, opened 2023-02-27
-Adds photo compression to viewmodel in order to notify user if photo is too large

andrewhamilton0:
-Adds photo compression to viewmodel in order to notify user if photo is too large
-Attempts to save photos to internal storage


app/src/main/java/com/andrew/tasky/agenda/data/util/BitmapConverters.kt
```diff
@@ -0,0 +1,27 @@
...
+            if (quality == 0) {
+                return null
+            }
+        }
+        return result
+    }
+
+    fun byteArrayToBitmap(data: ByteArray): Bitmap {
+        return BitmapFactory.decodeByteArray(data, 0, data.size)
+    }
```
philipplackner, line 26: Both these should be suspend functions

app/src/main/java/com/andrew/tasky/agenda/data/storage/InternalStorage.kt
```diff
@@ -0,0 +1,30 @@
+package com.andrew.tasky.agenda.data.storage
+
+import android.content.Context
+import com.andrew.tasky.agenda.data.event.photo.LocalPhotoDto
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import okio.IOException
+
+class InternalStorage(private val context: Context) {
```
philipplackner, line 9: You'll also need a function to delete an image after it has been uploaded to the server

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -88,11 +90,16 @@ class EventDetailViewModel @Inject constructor(
...
+    private val photos = _photo.asStateFlow()
     fun addPhoto(uri: Uri) {
-        val photo = EventPhoto.Local(uri = uri.toString())
-        _photo.value += photo
+        viewModelScope.launch {
+            val byteArray = uriByteConverter.uriToByteArray(uri = uri)
+            val photo = UiEventPhoto.LocalPhoto(
+                bitmap = BitmapConverters.byteArrayToBitmap(byteArray)
+            )
+            _photo.value += photo
```
philipplackner, line 101: Should be `_photos`

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -254,11 +309,34 @@ class EventDetailViewModel @Inject constructor(
...
+                internalStoragePhotos.forEach { internalStoragePhoto ->
+                    _photo.update { list ->
+                        list.toMutableList().plus(
+                            UiEventPhoto.LocalPhoto(
+                                key = internalStoragePhoto.key,
+                                bitmap = BitmapConverters.byteArrayToBitmap(
+                                    internalStoragePhoto.byteArray
+                                )
+                            )
+                        )
```
philipplackner, line 336: No need to convert to mutable list before, you can also add elements to an immutable one (it will create a new list under the hood):
```
_photos.update { list -> 
    list + UiEventPhoto.LocalPhoto(
                    key = internalStoragePhoto.key,
                    bitmap = BitmapConverters.byteArrayToBitmap(
                        internalStoragePhoto.byteArray
                    )
                )
}
```

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -254,11 +309,34 @@ class EventDetailViewModel @Inject constructor(
...
         }
         savedStateHandle.get<Boolean>("isInEditMode")?.let { initialEditMode ->
             setEditMode(initialEditMode)
         }
+
+        val keysList = savedStateHandle.get<AgendaItem.Event>("event")?.photos?.map { it.key }
+        viewModelScope.launch {
+            if (keysList != null) {
+                val internalStoragePhotos = repository.getLocalPhotos(keysList)
+                internalStoragePhotos.forEach { internalStoragePhoto ->
```
philipplackner, line 327: Why not simply something like 
```
_photos.update {
    repository.getLocalPhotos(keysList)
}
```
instead of all what you have?

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -197,39 +202,89 @@ class EventDetailViewModel @Inject constructor(
...
+                                } else {
+                                    EventPhoto.Local(key = eventPhoto.key)
+                                }
+                            }
+                            is UiEventPhoto.RemotePhoto -> {
+                                eventPhoto.remoteEventPhoto
+                            }
+                            else -> {
+                                null
+                            }
```
philipplackner, line 247: I'd avoid putting side effects in such lambdas like map, filter, etc. A side effect in this case would be something that's not related to mapping, like compressing & saving the local photo. You can quickly miss that this mapper actually does that and it makes the code quite unreadable.
Instead, after mapping, you can simply run a foreach loop in which you compress and save all images, so that these are separate actions.
Also, why do you always save all photos in the cache? Don't you only want to save them if uploading the event failed (due to missing internet for example)? I'd do that in the repository therefore where you know if that succeeded or not (or in the worker, if you use that).

philipplackner (approved):

PR #67, opened 2023-03-09
Moves photo compression to data layer after api upload failure

andrewhamilton0:


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -59,9 +59,8 @@ class AgendaRepositoryImpl(
             startEpochMilli = startEpochMilli,
             endEpochMilli = endEpochMilli
         ).map {
-            it.map {
-                eventEntity ->
-                eventEntity.toEvent()
+            it.map { eventEntity ->
+                eventEntity.toEvent(appContext)
```
philipplackner, line 63: Every time, you're executing a suspend function in a loop (which the `map` function is), you can make a mistake that really harms the performance. Here, you're executing the mapper sequentially, that means the function will first map event1, when that's done it'll map event2, then event3, and so on.
However, mapping an event isn't dependent on mapping a previous event. So you could execute all these `toEvent()` functions in parallel rather than sequentially.
If we assume that executing `toEvent()` takes 100ms, with your approach this function would take 1000ms.
If you execute all in parallel though, the function would only take 100ms. And the larger your list gets, the bigger the performance problem gets. 

So here's how you can fix this:
```
val mappedEvents = supervisorScope {
    eventEntities.map { entity ->
        async {
            entity.toEvent(appContext)
        }
    }
    .map { it.await() }
}
```

With `supervisorScope`, you can get a reference to a coroutine scope inside a suspend function (in this case inside `getAgendaItem()`. That's needed to launch independent coroutines which run in parallel. `supervisorScope` means if one of these coroutines launched in it will fail, the others will still run. There's also `coroutineScope` which will fail all coroutines if just one of them fails. 
With `async`, we launch a new independent coroutine which holds a result. The result can then be retrieved with `await()`. Hope that's all clear, if not just ask 👍

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -1,33 +1,141 @@
...
+                is Resource.Error -> {
+                    db.getEventDao().upsertModifiedEvent(
+                        ModifiedEventEntity(
+                            id = compressedEvent.id,
+                            modifiedType = ModifiedType.CREATE
+                        )
+                    )
+                    compressedEvent.photos.filterIsInstance<LocalPhoto>().forEach {
+                        saveLocalPhotoInternally(it)
+                        it.savedInternally = true
```
philipplackner, line 70: I'd never use `var` parameters for data classes to change the fields. That's very vulnerable for race conditions. Instead, make it `val` and use `copy()`:
```
saveLocalPhotoInternally(it.copy(savedInternally = true))
```

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -1,33 +1,141 @@
...
+                    }
+                    UpsertEventResult(photosTooBig = photosDeleted, resource = Resource.Error())
```
philipplackner, line 72: I'm not sure if I understand why the deleted photos are the too big photos here? 🤔

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -1,33 +1,141 @@
...
+                            name = "update_event_request",
+                            value = Json.encodeToString(compressedEvent.toUpdateEventRequest())
+                        ),
+                    photoData = compressedEvent.photos.filterIsInstance<LocalPhoto>().mapIndexed {
+                        index, eventPhoto ->
+                        MultipartBody.Part
+                            .createFormData(
+                                name = "photo$index",
+                                filename = eventPhoto.key,
+                                body = eventPhoto.byteArray!!.toRequestBody()
```
philipplackner, line 100: Are you 100% sure `byteArray` will never be null here? A more careful approach would be this:
```
body = eventPhoto.byteArray?.toRequestBody() ?: return@mapIndexed
```
That way you just ignore it if it doesn't exist.

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -1,33 +1,141 @@
...
+                            it.toEventEntity(
+                                isDone = compressedEvent.isDone,
+                                isGoing = compressedEvent.isGoing
+                            )
+                        )
+                    }
+                    compressedEvent.photos.filterIsInstance<LocalPhoto>().filter {
+                        it.savedInternally
+                    }.forEach {
+                        ImageStorage(context).deleteImage(it.key)
```
philipplackner, line 133: You could create a global reference of your `ImageStorage`, so you don't need to recreate it on every loop iteration

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -68,29 +176,58 @@ class EventRepositoryImpl @Inject constructor(
...
-            keysWithJpgEnding.contains(it.key)
-        }
+    private data class CompressEventResult(val event: AgendaItem.Event, val photosDeleted: Int)
+    private suspend fun compressEvent(event: AgendaItem.Event): CompressEventResult {
+        var photosDeleted = 0
+        val compressedEvent = event.copy(
+            photos = event.photos.mapNotNull { eventPhoto ->
+                when (eventPhoto) {
+                    is LocalPhoto -> {
+                        when (eventPhoto.savedInternally) {
```
philipplackner, line 198: If there are only 2 branches, I'd use if/else instead

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -68,29 +176,58 @@ class EventRepositoryImpl @Inject constructor(
...
+                            true -> {
+                                val byteArray = ImageStorage(context).getByteArray(eventPhoto.key)
+                                eventPhoto.copy(byteArray = byteArray)
+                            }
+                            false -> {
+                                val byteArray = eventPhoto.bitmap?.let { bitmap ->
+                                    BitmapConverters.bitmapToCompressByteArray(
+                                        bitmap, 1000000
```
philipplackner, line 206: `1000000` looks a bit like a magic number, you don't know at the first glance what it stands for. I'd either use named parameters here or make it a constant, for example `MAX_PHOTO_SIZE_IN_BYTES`

app/src/main/java/com/andrew/tasky/agenda/data/storage/ImageStorage.kt
```diff
@@ -0,0 +1,63 @@
...
+            val files = context.filesDir.listFiles()
+            files?.filter { it.canRead() && it.isFile && it.name.endsWith(".jpg") }?.map {
+                EventPhoto.Local(
+                    key = it.name.removeSuffix(".jpg"),
+                    bitmap = BitmapConverters.byteArrayToBitmap(it.readBytes()),
+                    savedInternally = true
+                )
+            } ?: emptyList()
+        }
+    }
```
philipplackner, line 54: When do you need to load all images for all events? Don't you rather only need to load them for a specific event?

app/src/main/java/com/andrew/tasky/agenda/data/storage/ImageStorage.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    suspend fun deleteImage(photoKey: String): Boolean? {
```
philipplackner, line 56: I wouldn't return nullable booleans, that makes things quite unclear since the boolean gets a 3rd state when it's null. From the outside what is the difference between the boolean being null or false? There is none.
You can return a non-null boolean with a comparison:
```files?.find { it.name == filename }?.delete() == true```
If that's null, the comparison will be false and therefore false will be returned

File: app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt [hunk 10]
```diff
@@ -255,33 +211,35 @@ class EventDetailViewModel @Inject constructor(
...
+                selectedEndDate.value,
+                selectedEndTime.value
+            ),
+            reminderTime = selectedReminderTime.value,
+            photos = photos.value,
+            attendees = attendees.value,
+            isCreator = savedStateHandle.get<AgendaItem.Event>("event")?.isCreator ?: true,
+            host = savedStateHandle.get<AgendaItem.Event>("event")?.host,
+            deletedPhotoKeys = emptyList(), // TODO setup deletedPhotosKeys
+            isGoing = savedStateHandle.get<AgendaItem.Event>("event")?.isGoing ?: true
```
philipplackner, line 239: I'd avoid passing the event as a `Parcelable` to the event detail screen. With that, you're violating the single source of truth principle. In your app, the single source is your local DB. If you load from multiple sources (here from savedStateHandle for example), you're likely going to run into bugs. Imagine, your AgendaScreen didn't fully update its items yet, but there was a change in your DB for the event the user clicks on. Then if you click on the event on the AgendaScreen (which is outdated), you'll get an outdated detail view and if the user then saves it, it'll override the new event with the outdated one. 
Therefore, for detail screens in most cases I'd just pass the ID of the event and load it from the DB again.

philipplackner (approved):

PR #68, opened 2023-03-12
Sends only agenda item Id through fragment

andrewhamilton0:
-Sends only agenda item Id through fragment and receives the rest of the info from local database
-Maps event suspend function with supervisor scope and async
-Makes global reference to ImageStorage in EventRepositoryImpl.kt
-Deletes load all images

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailFragment.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailFragment.kt
@@ -63,8 +63,8 @@ class EventDetailFragment : Fragment(R.layout.fragment_event_detail) {
                 viewModel.setEditMode(true)
             }
             header.saveButton.setOnClickListener {
+                progressBar.isVisible = true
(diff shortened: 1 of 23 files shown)


app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailFragment.kt
```diff
@@ -63,8 +63,8 @@ class EventDetailFragment : Fragment(R.layout.fragment_event_detail) {
                 viewModel.setEditMode(true)
             }
             header.saveButton.setOnClickListener {
+                progressBar.isVisible = true
```
philipplackner, line 66: I'd make all your view changes dependent on a state from the ViewModel, otherwise they won't survive config changes like screen rotations. In this case, your `saveEvent()` function from the ViewModel should change an `isSavingEvent` boolean which your fragment observes as a flow or so.
With your approach if you hit save and rotate the device while saving the loading spinner will disappear.

philipplackner (approved):

PR #70, opened 2023-03-29
Changes DeleteConfirmation to use string resource instead of agenda type

andrewhamilton0:
-Changes DeleteConfirmation to use string resource instead of agenda type
-Deletes Agenda Type
-Fixes item_agenda.xml to be expandable
-Moves EventCompression to separate class from event repo
-Uses event repo as parameter in EventEntity.toEvent mapper
-Moves get photo from mapper to event repo

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/event/EventCompressor.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/event/EventCompressor.kt
@@ -1,0 +1,61 @@
+package com.andrew.tasky.agenda.data.event
+
+import android.content.Context
+import com.andrew.tasky.agenda.data.storage.ImageStorage
+import com.andrew.tasky.agenda.data.util.BitmapConverters
+import com.andrew.tasky.agenda.domain.models.AgendaItem
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.supervisorScope
+import kotlinx.coroutines.withContext
+
+class EventCompressor(context: Context) {
+
+    private val imageStorage = ImageStorage(context)
+    private val bitmapConverter = BitmapConverters
+
+    data class CompressEventResult(val event: AgendaItem.Event, val photosDeleted: Int)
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/event/EventCompressor.kt
```diff
@@ -0,0 +1,61 @@
...
+import android.content.Context
+import com.andrew.tasky.agenda.data.storage.ImageStorage
+import com.andrew.tasky.agenda.data.util.BitmapConverters
+import com.andrew.tasky.agenda.domain.models.AgendaItem
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.supervisorScope
+import kotlinx.coroutines.withContext
+
+class EventCompressor(context: Context) {
```
philipplackner, line 12: Aren't it rather the event's photos that are compressed and not the event itself?

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -199,44 +198,16 @@ class EventRepositoryImpl @Inject constructor(
...
-                        }.flatten()
+    override suspend fun getLocalPhotos(localPhotoKeys: List<String>): List<EventPhoto.Local> {
+        return withContext(Dispatchers.IO) {
+            supervisorScope {
+                async {
+                    localPhotoKeys.map { key ->
+                        val bitmap = imageStorage.getBitmap(key)
+                        EventPhoto.Local(key = key, bitmap = bitmap, savedInternally = true)
+                    }
                 }
```
philipplackner, line 209: Wrong place for async, this will execute the whole mapping inside async and therefore still load the bitmaps sequentially.
Instead, the async block should be inside the map block so an independent coroutine is launched for every call to `getBitmap()`

philipplackner (approved):

PR #71, opened 2023-04-02
Sets up Agenda Notification Channels and Service

andrewhamilton0:
Sets up Agenda Notification Channels and Service

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/presentation/notifications/AgendaNotificationService.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/presentation/notifications/AgendaNotificationService.kt
@@ -1,0 +1,49 @@
+package com.andrew.tasky.agenda.presentation.notifications
+
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.content.Context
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/andrew/tasky/agenda/presentation/notifications/AgendaNotificationService.kt
```diff
@@ -0,0 +1,49 @@
...
+            context, 0, intent,
+            PendingIntent.FLAG_IMMUTABLE
+        )
+        val notification = NotificationCompat.Builder(context, EVENT_CHANNEL_ID)
+            .setSmallIcon(R.drawable.ic_notification_agenda)
+            .setContentTitle(title)
+            .setContentText(description)
+            .setContentIntent(pendingIntent)
+
+        notificationManager.notify(1, notification.build())
```
philipplackner, line 38: If you call notify with the same ID again, it will replace an existing notification. Also, here it would never show 2+ notifications at once. The ID needs to be unique per notification (unless you want to update an existing one)

philipplackner (approved):

PR #73, opened 2023-04-03
Refactors code EventRepositoryImpl.kt

andrewhamilton0:
-Refactors code in upsertEvent function in EventRepositoryImpl.kt to be smaller and easier to read


app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -32,116 +33,199 @@ class EventRepositoryImpl @Inject constructor(
...
-                        )
                     }
-                    compressedEvent.photos.filterIsInstance<LocalPhoto>().filter {
-                        it.savedInternally
-                    }.forEach {
-                        imageStorage.deleteImage(it.key)
+            )
+        }
+        return result
+    }
```
philipplackner, line 190: If all you do is create a variable and then return it without using it for anything, I'd just return the value right away without creating the variable :)

philipplackner (approved):

PR #74, opened 2023-04-14
-Sets up notification broadcast receiver and scheduler

andrewhamilton0:


app/src/main/java/com/andrew/tasky/agenda/data/agenda/notifications/AgendaNotificationService.kt
```diff
@@ -0,0 +1,45 @@
...
+    private val context: Context
+) {
+
+    private val notificationManager = context
+        .getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+
+    fun showNotification(info: AgendaNotificationInfo) {
+        val navArs = Bundle().apply {
+            putString("id", info.id)
+            putBoolean("isInEditMode", false)
```
philipplackner, line 21: Couldn't you work with default args here, so you wouldn't need to provide this?

app/src/main/java/com/andrew/tasky/agenda/data/agenda/workManagers/AgendaNotificationScheduleWorker.kt
```diff
@@ -0,0 +1,41 @@
...
+        items.forEach { item ->
+            when (item) {
+                is AgendaItem.Event -> scheduler.schedule(item.toNotificationInfo())
+                is AgendaItem.Reminder -> scheduler.schedule(item.toNotificationInfo())
+                is AgendaItem.Task -> scheduler.schedule(item.toNotificationInfo())
+            }
+        }
+        return Result.success()
+    }
+}
```
philipplackner, line 41: There's no real reason to do this in a worker. WorkManager is used to execute longer running tasks reliably (such as uploading an event with photos), but scheduling an alarm is pretty much instant, so you could do that as soon as the user saves the agenda item.

app/src/main/java/com/andrew/tasky/agenda/domain/AgendaNotificationScheduler.kt
```diff
@@ -0,0 +1,8 @@
+package com.andrew.tasky.agenda.domain
+
+import com.andrew.tasky.agenda.domain.models.AgendaNotificationInfo
+
+interface AgendaNotificationScheduler {
+    fun schedule(agendaNotificationInfo: AgendaNotificationInfo)
+    fun cancel(agendaNotificationInfo: AgendaNotificationInfo)
```
philipplackner, line 7: Wouldn't it be enough to just pass the ID of the agenda item here? That's all you need to get its hashcode and cancel the notification.

PR #75, opened 2023-04-14
-Sets up notification broadcast receiver and scheduler

andrewhamilton0:


andrewhamilton0:
Moved some files around for refactoring and also changed photo compression function to  compress unsaved photos only

andrewhamilton0:
Accidentally named commit be596ab "Sets up notification broadcast and scheduler" which is innaccurate and is on previous branch. This commit is just moving files to more appropriate architecture packages

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -36,7 +37,27 @@ class EventRepositoryImpl @Inject constructor(
     private val eventPhotoCompressor = EventPhotoCompressor()
 
     override suspend fun upsertEvent(event: Event): UpsertEventResult {
-        val (newEvent, deletedPhotoCount) = compressAndDeletePhotosFromEvent(event)
+        val (compressedPhotos, deletedPhotoCount) = compressAndDeletePhotos(
+            event.photos.filterIsInstance<LocalPhoto>().filter {
+                !it.savedInternally
+            }
```
philipplackner, line 43: Here you could use a named parameter to make it clear what this lambda evaluates to. That just reduces the time needed to think what this is for :)

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -36,7 +37,27 @@ class EventRepositoryImpl @Inject constructor(
...
+        )
+        val eventWithCompressedPhotos = event.copy(
+            photos = event.photos.filterNot { photo ->
+                photo is LocalPhoto && !photo.savedInternally
+            }.toMutableList().apply {
+                compressedPhotos.forEach { compressedPhoto ->
+                    this.add(compressedPhoto)
+                }
+            }.toImmutableList()
```
philipplackner, line 52: `toImmutableList()` isn't an official Kotlin list type (you can see in the imports it's some internal OkHttp stuff). You can just use `toList()` here

app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailViewModel.kt
```diff
@@ -29,9 +31,13 @@ import kotlinx.coroutines.withContext
 class EventDetailViewModel @Inject constructor(
     private val savedStateHandle: SavedStateHandle,
     private val repository: EventRepository,
-    private val uriByteConverter: UriByteConverter
+    private val uriByteConverter: UriByteConverter,
+    private val prefs: SharedPreferences
```
philipplackner, line 35: Here, you're binding a data layer dependency to your presentation layer. Instead, I'd recommend to abstract out this preferences implementation in domain and data and then use the domain interface in your ViewModel

PR #76, opened 2023-04-14
Deleting remote photos locally and without network, now uploads delet…

andrewhamilton0:
Deleting remote photos locally and without network, now uploads deleted photos correctly


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -75,6 +72,46 @@ class AgendaRepositoryImpl(
         }
     }
 
+    override suspend fun getOneTimeAgendaItemsBetweenTimes(
+        startEpochMilli: Long,
+        endEpochMilli: Long
+    ): List<AgendaItem> {
```
philipplackner, line 78: Wouldn't it make more sense to also return a flow here? Then you still have the option to get them once by using `first()` on that returned flow, but you're also more flexible with listening to changes. Considering all your DAO functions return flows, that seems to be a simple change, since you just need to combine them

philipplackner (approved):

PR #77, opened 2023-04-19
-Makes a shared prefs interface and impl

andrewhamilton0:
-Deletes an unused function from AgendaRepositoryImpl.kt
 -Delete AgendaNotificationScheduleWorker.kt
-Gets agenda item info from broadcast receiver
-Changes agenda item notification scheduler to only take reminder time and ID
-Uses default arg for IsInEditMode in AgendaNotificationService.kt


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -72,44 +72,14 @@ class AgendaRepositoryImpl(
...
-
-        return agendaItems.sortedBy { it.startDateAndTime }
+    override suspend fun getAgendaItemById(id: String): AgendaItem? {
+        val event = eventRepository.getEvent(id)
+        val task = taskRepository.getTask(id)
+        val reminder = reminderRepository.getReminder(id)
+
+        val agendaItems = listOf(event, task, reminder)
+        agendaItems.forEach { if (it != null) return it }
+        return null
```
philipplackner, line 82: You could simply do this:
```
return agendaItems.first { it != null }
```

app/src/main/java/com/andrew/tasky/agenda/data/agenda/notifications/AgendaBroadcastReceiver.kt
```diff
@@ -3,17 +3,36 @@ package com.andrew.tasky.agenda.data.agenda.notifications
...
 
+    @Inject
+    lateinit var agendaRepository: AgendaRepository
+
     override fun onReceive(context: Context?, intent: Intent?) {
-        val info = intent?.getParcelableExtra(
-            AGENDA_NOTIF_SCHED_INTENT,
-            AgendaNotificationInfo::class.java
-        )
+        val coroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
philipplackner, line 24: The default dispatcher is rather used for CPU heavy things which isn't the case here. I'd simply use `Dispatchers.Main`

app/src/main/java/com/andrew/tasky/agenda/data/agenda/notifications/AgendaBroadcastReceiver.kt
```diff
@@ -3,17 +3,36 @@ package com.andrew.tasky.agenda.data.agenda.notifications
...
+        val id = intent?.getStringExtra(AGENDA_NOTIF_SCHED_INTENT)
         val service = context?.let { AgendaNotificationService(it) }
-        info?.let { service?.showNotification(it) }
+
+        coroutineScope.launch {
+            val item = id?.let { agendaRepository.getAgendaItemById(it) }
+            when (item) {
+                is AgendaItem.Event -> service?.showNotification(item.toNotificationInfo())
+                is AgendaItem.Reminder -> service?.showNotification(item.toNotificationInfo())
+                is AgendaItem.Task -> service?.showNotification(item.toNotificationInfo())
```
philipplackner, line 33: Couldn't you use a shared `toNotificationInfo()` function here, so you wouldn't need the when expression? You use an AgendaItem sealed class I assume, so you could put an abstract function `toNotificationInfo()` in there and implement it in each of the sub classes

app/src/main/java/com/andrew/tasky/agenda/data/event/EventRepositoryImpl.kt
```diff
@@ -308,6 +311,22 @@ class EventRepositoryImpl @Inject constructor(
...
+            agendaId = event.id,
+            time = ReminderTimeConversion.toZonedEpochMilli(
+                startLocalDateTime = event.startDateAndTime,
+                reminderTime = event.reminderTime
+            )
+        )
+    }
+
+    private suspend fun cancelScheduledNotification(eventId: String) {
+        db.getEventDao().getEventById(eventId)?.toEvent(this)?.let {
```
philipplackner, line 325: Should be fine to just call `cancel` with the passed ID here. If there's no scheduled job it will simply ignore it

app/src/main/java/com/andrew/tasky/core/data/SharedPrefsImpl.kt
```diff
@@ -0,0 +1,41 @@
+package com.andrew.tasky.core.data
+
+import android.content.SharedPreferences
+import com.andrew.tasky.core.domain.SharedPrefs
+
+class SharedPrefsImpl(
+    private val sharedPrefs: SharedPreferences
+) : SharedPrefs {
+
+    override fun matchesUserId(s: String): Boolean {
```
philipplackner, line 10: I'd avoid using such abbreviations. What is `s`? I assume it's just a user ID? Then I'd call it like that.

philipplackner (approved):

PR #78, opened 2023-04-22
checks if token is valid or offline

andrewhamilton0:
-Checks if token is valid or offline
-Makes agenda notification mapper into one mapper
-Sets default nav map values for agenda items


andrewhamilton0:
Could not figure out how to remove previous code from this branch, so I'm closing this one and starting a new one

PR #79, opened 2023-04-22
(1) Checks if token is valid or online

andrewhamilton0:
-checks if token is valid or online
-Makes agenda notification mapper into one mapper
-Sets default nav map values for agenda items

ALSO this PR has so many commits because I was attempting to remove the last code from the last PR/Branch from this branch because I forgot to merge before uploading this. I was unsuccessful so I made a new branch


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaMappers.kt
```diff
@@ -0,0 +1,51 @@
...
+            AgendaNotificationInfo(
+                title = title,
+                description = description,
+                id = id,
+                notificationChannel = AgendaNotificationService.TASK_CHANNEL_ID,
+                navDestinationId = R.id.taskDetailFragment,
+                notificationZonedMilliTime = ReminderTimeConversion.toZonedEpochMilli(
+                    startDateAndTime,
+                    reminderTime
+                )
```
philipplackner, line 47: Here only 2 fields really change, `navDestinationId` and `notificationChannel`. You could also instantiate only these 2 fields in the when expression:
```
val (channel, navId) = when(this) {
    is AgendaItem.Event -> AgendaNotificationService.EVENT_CHANNEL_ID to R.id.event_nav
    is AgendaItem.Task -> AgendaNotificationService.TASK_CHANNEL_ID to R.id.task_nav
    is AgendaItem.Reminder -> AgendaNotificationService.REMINDER_CHANNEL_ID to R.id.reminder_nav
}
```
And then only create one `AgendaNotificationInfo`

app/src/main/java/com/andrew/tasky/auth/data/AuthRepositoryImpl.kt
```diff
@@ -85,4 +86,14 @@ class AuthRepositoryImpl(
         prefs.clearPrefs()
         agendaRepository.deleteAllAgendaTables()
     }
+
+    override suspend fun isAuthorizedToLogin(): Boolean {
+        val result = authenticate()
+        when (result) {
+            is Resource.Error -> {
+                return result.errorType is ApiErrorType.IOException && prefs.containsJwt()
```
philipplackner, line 94: Actually, here you only want to check for HTTP 401 specifically which means not authorized. For other HTTP status codes such as 500 (when there's an error server-side which is not the client's fault), you don't want to treat this as unauthorized

philipplackner (approved):

PR #81, opened 2023-05-13
Adds persistent notifications

andrewhamilton0:
-Adds persistent notifications
-Fixes notification glitch where notification doesn't open detail screen 
-Adds where creator is editing determines PhotoDetailFragment.kt delete button

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
@@ -4,6 +4,7 @@ import android.app.AlarmManager
 import android.content.Context
 import android.icu.util.TimeZone
 import android.util.Log
+import com.andrew.tasky.agenda.data.agenda.notifications.PersistedNotifEntity
 import com.andrew.tasky.agenda.data.database.AgendaDatabase
 import com.andrew.tasky.agenda.data.event.toEvent
 import com.andrew.tasky.agenda.data.event.toEventEntity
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -295,6 +352,30 @@ class AgendaRepositoryImpl(
...
+
+    override suspend fun upsertPersistedNotification(agendaId: String) {
+        db.getPersistedNotifDao().upsertPersistedNotif(
+            PersistedNotifEntity(agendaId = agendaId)
+        )
+    }
+
+    override suspend fun deleteAllPersistedNotifs() {
+        db.getPersistedNotifDao().deleteAllPersistedNotifs()
+    }
```
philipplackner, line 377: I wonder why you need to save these notifications persistently? If you want to reschedule them after a device reboot, couldn't you just add a Room query to retrieve all future agenda items and use their times to schedule the notifications?

app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -295,6 +352,30 @@ class AgendaRepositoryImpl(
         cancelAllNotifications()
     }
 
+    override suspend fun sendPersistedNotifications() {
+        // Adds ten seconds because scheduler does not schedule alarms even 1 second old
```
philipplackner, line 356: Great use of a comment 💪

app/src/main/java/com/andrew/tasky/agenda/data/agenda/notifications/BootupNotificationBroadcastReceiver.kt
```diff
@@ -0,0 +1,28 @@
...
+
+@AndroidEntryPoint
+class BootupNotificationBroadcastReceiver : BroadcastReceiver() {
+
+    @Inject
+    lateinit var agendaRepository: AgendaRepository
+
+    override fun onReceive(context: Context?, intent: Intent?) {
+        if (intent?.action == Intent.ACTION_BOOT_COMPLETED) {
+            val coroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Main)
```
philipplackner, line 22: The supervisor job isn't really necessary here, since you only launch a single coroutine from this scope. Supervisor job means that if one job fails, the other jobs launched in this scope will continue running, but since you only have a single job here it's redundant

app/src/main/java/com/andrew/tasky/core/presentation/main/MainActivity.kt
```diff
@@ -24,11 +25,14 @@ class MainActivity : AppCompatActivity() {
...
-                if (viewModel.isUserLoggedIn.value) {
-                    navController.navigate(R.id.action_global_agendaFragment)
-                }
                 viewModel.isLoading.value
             }
         }
+
+        collectLatestLifecycleFlow(viewModel.isUserLoggedIn) { isLoggedIn ->
+            if (!isLoggedIn) {
+                navController.navigate(R.id.action_global_loginFragment)
```
philipplackner, line 34: Here you might want to also clear the backstack, since otherwise the user could press back on the login screen and get back to the agenda screen if this gets triggered from there

PR #82, opened 2023-05-13
Made DateTimeConversion.kt and ReminderTimeConversion.kt interface classes

andrewhamilton0:
-Made DateTimeConversion.kt and ReminderTimeConversion.kt interface classes and put implementations in data layer

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderRepositoryImpl.kt
@@ -5,6 +5,7 @@ import android.util.Log
 import com.andrew.tasky.agenda.data.database.AgendaDatabase
 import com.andrew.tasky.agenda.data.util.ModifiedType
 import com.andrew.tasky.agenda.domain.AgendaNotificationScheduler
+import com.andrew.tasky.agenda.domain.DateTimeConversion
 import com.andrew.tasky.agenda.domain.ReminderRepository
 import com.andrew.tasky.agenda.domain.ReminderTimeConversion
 import com.andrew.tasky.agenda.domain.models.AgendaItem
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/reminder/ReminderRepositoryImpl.kt
```diff
@@ -111,12 +141,20 @@ class ReminderRepositoryImpl @Inject constructor(
         }
     }
 
+    override suspend fun completeReminderNotification(reminderId: String) {
+        val reminder = db.getReminderDao().getReminderById(
+            reminderId
+        )?.copy(notificationCompleted = true)
```
philipplackner, line 147: Also this shouldn't be necessary since you know that a notification must be completed if `current time > agenda item time - remind at time`

app/src/main/java/com/andrew/tasky/agenda/data/util/DateTimeConversionImpl.kt
```diff
@@ -0,0 +1,25 @@
...
+        ).toInstant().toEpochMilli()
+    }
+
+    override fun zonedEpochMilliToLocalDateTime(zonedEpochMilli: Long): LocalDateTime {
+        return ZonedDateTime.ofInstant(
+            Instant.ofEpochMilli(zonedEpochMilli),
+            ZoneId.of(TimeZone.getDefault().id)
+        ).toLocalDateTime()
+    }
+}
```
philipplackner, line 25: For this you don't need to put the implementation in data, since there's nothing Android related about `ZonedDateTime`. If something comes from Android or just Java/Kotlin, you can easily see at the imports. However, what would make this Android related is if you use the Android specific `TimeZone` import (there are two - one from java and one from Android).
The difference is that the one from Android will consider the time zone the user set in their Android settings which is why I'd recommend that one for apps, otherwise it might lead to incorrect times. In that case, I'd only put that time zone specific implementation in data, but leave the conversion up to domain. For example like this:
```
// domain
interface TimeZoneProvider {
    fun getTimeZone(timeZoneId: String?): ZoneId
}

// data
import android.icu.util.TimeZone

object AndroidTimeZoneProvider : TimeZoneProvider {
    override fun getTimeZone(timeZoneId: String?): ZoneId {
        // Check if zone ID is supported
        val adjustedTimeZone = if (timeZoneId in TimeZone.getAvailableIDs()) {
            timeZoneId
        } else TimeZone.getDefault().id // if not, fallback to default time zone

        return ZoneId.of(adjustedTimeZone)
    }
}

// Time zone conversion in domain
class DateTimeConversion(private val timeZoneProvider: TimeZoneProvider) {
    
    // Your conversion functions here
} 
```

PR #84, opened 2023-05-19
-Adds crashlytics

andrewhamilton0:
-Adds async functions to AgendaRepositoryImpl

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
@@ -212,54 +212,75 @@ class AgendaRepositoryImpl(
         taskRepository.uploadCreateAndUpdateModifiedTasks()
         eventRepository.uploadCreateAndUpdateModifiedEvents()
 
-        val reminderDeleteIds = db.getReminderDao().getModifiedReminders().first().filter {
-            it.modifiedType == ModifiedType.DELETE
-        }.map {
-            it.id
-        }
-        val taskDeleteIds = db.getTaskDao().getModifiedTasks().first().filter {
-            it.modifiedType == ModifiedType.DELETE
-        }.map {
-            it.id
-        }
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -212,54 +212,75 @@ class AgendaRepositoryImpl(
...
-            )
-            val results = getResourceResult { agendaApi.syncAgendaItems(syncAgendaRequest) }
-            when (results) {
-                is Resource.Error -> {
-                    Log.e(
-                        "SyncAgendaItem Error",
-                        results.message?.asString(appContext) ?: "Unknown Error"
-                    )
+        val coroutine = CoroutineScope(Dispatchers.IO + SupervisorJob())
+        return coroutine.async {
```
philipplackner, line 216: I'd avoid using custom scopes in components that don't have a lifecycle (such as this repo - it's a singleton). Instead I'd rely on using `viewModelScope` and making this a suspend function. You can then still use async this way:
```
supervisorScope {
    async { ... }
}

```

app/src/main/java/com/andrew/tasky/agenda/data/agenda/AgendaRepositoryImpl.kt
```diff
@@ -212,54 +212,75 @@ class AgendaRepositoryImpl(
...
+            if (reminderDeleteIds.isNotEmpty() ||
+                taskDeleteIds.isNotEmpty() ||
+                eventDeleteIds.isNotEmpty()
+            ) {
+                val syncAgendaRequest = SyncAgendaRequest(
+                    eventDeleteIds,
+                    taskDeleteIds,
+                    reminderDeleteIds
+                )
+                val results = getResourceResult { agendaApi.syncAgendaItems(syncAgendaRequest) }
```
philipplackner, line 251: Aren't you just syncing the locally deleted agenda items here? The function is called `syncModifiedAgendaItems`

PR #85, opened 2023-05-24
-Closes add attendee dialog on success

andrewhamilton0:
-Closes add attendee dialog on success
-Has initial date and time pickers picked based on initial state values
-Adds 30 minutes to start or end time if end time is less than start
-Added notification permission checker
-Fixed night mode theme to be disabled
-Changed way alarms are canceled

### DIFF
--- a/app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailFragment.kt
+++ b/app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailFragment.kt
@@ -1,11 +1,17 @@
 package com.andrew.tasky.agenda.presentation.screens.event_detail
 
+import android.Manifest
+import android.content.Context
+import android.content.pm.PackageManager
 import android.graphics.Paint
+import android.os.Build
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/andrew/tasky/agenda/presentation/screens/event_detail/EventDetailFragment.kt
```diff
@@ -525,4 +578,19 @@ class EventDetailFragment : Fragment(R.layout.fragment_event_detail) {
             EventDetailFragmentDirections.actionEventDetailFragmentToPhotoDetailFragment()
         )
     }
+
+    @RequiresApi(Build.VERSION_CODES.TIRAMISU)
+    private fun areNotificationPermissionsEnabled(context: Context): Boolean {
```
philipplackner, line 583: Wouldn't `isNotificationPermissionGranted` make more sense for the name, since it's not multiple permissions? (and because permissions aren't really enabled or disabled, but rather granted or not granted)

philipplackner (approved):

